PR #28, opened 2022-07-16
 Add additional send operation commands

jwmdev:
Added and tested Transactions, BuyRolls, and  SellRolls.


pkg/node/sendoperation/executesc/executesc.go
```diff
@@ -0,0 +1,72 @@
...
+	// Coins
+	n = binary.PutUvarint(buf, e.Coins)
+	msg = append(msg, buf[:n]...)
+
+	// GasPrice
+	n = binary.PutUvarint(buf, e.gasPrice)
+	msg = append(msg, buf[:n]...)
+
+	// data
+	msg = append(msg, e.data...)
```
gregLibert, line 69: It's not working because you need to set le length of the field. Hence the error message `Deserialization error: There is data left after operation deserialization` because the size of the field is longer than the remaining data.

```suggestion
        n = binary.PutUvarint(buf, uint64(len(e.data)))
        msg = append(msg, buf[:n]...)
      	msg = append(msg, e.data...)
```

jwmdev: > It's not working because you need to set le length of the field. Hence the error message `Deserialization error: There is data left after operation deserialization` because the size of the field is longer than the remaining data.

Fixed and now it is working... thanks for the guidance.

pkg/wallet/wallet.go
```diff
@@ -0,0 +1,81 @@
...
+	address := getAddressFromPubKey(pubKey)
+	w := &Wallet{
+		PrivateKey: privKey,
+		Address:    address,
+	}
+	return w, nil
+}
+
+//NewFromSeed expects base58 encoded secreat key as per massa and returns the wallet
+func NewFromSeed(key string) (*Wallet, error) {
```
gregLibert, line 35: I'm not sure to understand what is the purpose of this function other than reproductibility. If it's only for that, I don't understand its usage.

pkg/node/sendoperation/sendoperation.go
```diff
@@ -38,7 +38,7 @@ func (u JSONableSlice) MarshalJSON() ([]byte, error) {
 func message(expiry uint64, fee uint64, operation Operation) []byte {
 	msg := make([]byte, 0)
 	buf := make([]byte, binary.MaxVarintLen64)
-
+	fmt.Print("here")
```
gregLibert, line 41: This should be removed

pkg/node/sendoperation/transaction/transaction.go
```diff
@@ -0,0 +1,61 @@
...
+
+type Transaction struct {
+	recepientAddress string
+	amount           uint64
+}
+
+func New(recepientAddress string, amount uint64) *Transaction {
+	return &Transaction{
+		recepientAddress: recepientAddress,
+		amount:           amount * 1e9,
```
gregLibert, line 27: Why are you multiplying ?

pkg/node/sendoperation/transaction/transaction.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+type Transaction struct {
+	recepientAddress string
+	amount           uint64
+}
+
+func New(recepientAddress string, amount uint64) *Transaction {
+	return &Transaction{
+		recepientAddress: recepientAddress,
```
gregLibert, line 26: recepient => recipient

pkg/node/sendoperation/transaction/transaction.go
```diff
@@ -0,0 +1,61 @@
...
+		amount:           amount * 1e9,
+	}
+}
+
+func (t *Transaction) Content() interface{} {
+	addr, _, _ := base58.VersionedCheckDecode(t.recepientAddress[1:])
```
gregLibert, line 32: You should handle the error or t.recipientAddress shoul already be a byte array

pkg/wallet/wallet.go
```diff
@@ -0,0 +1,81 @@
+package wallet
```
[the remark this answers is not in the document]
gregLibert: You can now use the wallet from pkg/wallet

pkg/node/sendoperation/transaction/transaction.go
jwmdev: > Why are you multiplying ?

Multiplying because if you intent to send 1 MAS, the network will actually send 1e-9.

pkg/wallet/wallet.go
```diff
@@ -9,7 +9,7 @@ import (
 	"errors"
 
 	"github.com/massalabs/thyra/pkg/node/base58"
-	"golang.org/x/crypto/pbkdf2"
+	"github.com/xdg-go/pbkdf2"
```
gregLibert, line 12: @jwmdev why changing pbkdf2 function package ?

jwmdev: > @jwmdev why changing pbkdf2 function package ?

No reasons, I don't know what happened. Anyway, I see it resolved.

PR #61, opened 2022-08-22
Default wallet

Trottoir:
Wallet separation
Default wallet management for operations
DNS modification to retrieve domain names from SC and not locally anymore


pkg/my/domain.go
```diff
@@ -18,40 +21,51 @@ type Domains struct {
 	domains []Domain
 }
 
-func (d *Domains) List() []Domain {
-	return d.domains
-}
-
-func (d *Domains) Save() error {
-	cnt, err := json.Marshal(d.domains)
+func GetDomains(client *node.Client, nickname string) ([]string, error) {
```
gregLibert, line 24: Usage will be like : 
```go
...
my.GetDomains(...)
...
```
I porpose to change the function name to Domains to get :
```go
...
my.Domains(...)
...
```

Note: in Go, we avoid using Get prefix.

pkg/my/domain.go
```diff
@@ -18,40 +21,51 @@ type Domains struct {
...
+	if err != nil {
+		return nil, err
+	}
+	return domains, nil
 }
 
-func (d *Domains) Add(dom Domain) {
-	d.domains = append(d.domains, dom)
-}
+func GetOwnedWebsites(client *node.Client, domainNames []string) ([]*models.Websites, error) {
```
gregLibert, line 42: What is the model linking domain and website ?

I propose to rename the function to Website, that will allow an my.Websites usage.

pkg/node/getdatastoreentries.go
```diff
@@ -64,3 +64,48 @@ func DatastoreEntry(client *Client, address string, key string) (*DatastoreEntry
 
 	return &entry[0], nil
 }
+
+type GetDatastoreEntriesString struct {
+	Address string `json:"address"`
+	Key     string `json:"key"`
+}
+
+func DatastoreEntries(client *Client, params []GetDatastoreEntriesString) (*[]DatastoreEntryResponse, error) {
```
gregLibert, line 73: Does it make sens to remove DatastoreEntry or simply to change it content to DatastoreEntries[0] ?

pkg/onchain/dns/dns.go
```diff
@@ -10,16 +10,15 @@ import (
 	"github.com/massalabs/thyra/pkg/wallet"
 )
 
-const dnsRawAddress = "A1Q65NojVV5YPyZruVkeU1CGeS3tjLNwGSzAmZfAJPE5vuvus4C"
+const DnsRawAddress = "A12jkDPTcdhkqGg9VoKsTwvkBwZeSHQw7wJqQYKrNesKnjnGejuR"
```
gregLibert, line 13: Why changing it to a public value ?

pkg/sc/websiteStorer.go
```diff
@@ -1,7 +1,7 @@
 // GENERATED BY textFileToGoConst
 // GitHub:     github.com/logrusorgru/textFileToGoConst
-// input file: sc/websiteStorer.wasm
-// generated:  Tue Aug  9 17:21:32 CEST 2022
+// input file: sc\websiteStorer.wasm
```
gregLibert, line 3: painful. Maybe we can ask textFileToGoConst to avoid setting a date and the input file path to avoid having a diff ?

pkg/wallet/wallet.go
```diff
@@ -110,31 +110,52 @@ func FromYAML(raw []byte) (w Wallet, err error) {
...
+			wallets = append(wallets, wallet)
 		}
 	}
 
 	return wallets, nil
 }
 
-func FirstWallet() (Wallet, error) {
-	wallets, err := ReadWallets()
+func GetWallet(nickname string) (*Wallet, error) {
```
gregLibert, line 147: rename to Load

pkg/wallet/wallet.go
```diff
@@ -110,31 +110,52 @@ func FromYAML(raw []byte) (w Wallet, err error) {
 	return
 }
 
-func ReadWallets() (wallets []Wallet, e error) {
-	bytesInput, e := ioutil.ReadFile("wallet.json")
-	wallets = []Wallet{}
+func ReadWallets() ([]Wallet, error) {
```
gregLibert, line 113: is it returning all wallets ? In that case what about renaming it LoadAll ?

pkg/onchain/website/uploader.go
```diff
@@ -12,37 +12,26 @@ import (
...
 	// TODO should use a nickname argument
-	wallet, err := wallet.FirstWallet()
+	wallet, err := wallet.GetWallet(walletNickname)
 	if err != nil {
 		return "", err
 	}
 
 	// Prepare address to webstorage.
-	scAddress, err := onchain.DeploySC(client, wallet, []byte(sc.WebsiteStorer))
+	scAddress, err := onchain.DeploySC(client, *wallet, []byte(sc.WebsiteStorer))
```
gregLibert, line 25: Dereferencing the pointer makes the code look weird => this function should accept a pointer of wallet to avoir useless structure duplicate.

pkg/onchain/website/uploader.go
```diff
@@ -74,5 +63,5 @@ func Upload(atAddress string, content string) (string, error) {
 		return "", err
 	}
 
-	return onchain.CallFunction(client, wallet, addr, "initializeWebsite", param)
+	return onchain.CallFunction(client, *wallet, addr, "initializeWebsite", param)
```
gregLibert, line 66: Same as above

resource/api/swagger.yml
```diff
@@ -165,9 +165,46 @@ paths:
           description: Internal Server Error - The server has encountered a situation it does not know how to handle.
           schema:
             $ref: '#/definitions/Error'
-  /my/domains:
+    get:
+      operationId: mgmtWalletGetOne
```
gregLibert, line 169: mgmtWalletGetter

resource/api/swagger.yml
```diff
@@ -165,9 +165,46 @@ paths:
...
+          description: Unprocessable Entity - syntax is correct, but the server was unable to process the contained instructions.
+          schema:
+            $ref: '#/definitions/Error'
+        '500':
+          description: Internal Server Error - The server has encountered a situation it does not know how to handle.
+          schema:
+            $ref: '#/definitions/Error'
+  /my/domains/{nickname}:
     get:
       operationId: myDomains
```
gregLibert, line 201: myDomainsGetter

gregLibert (changes requested):

pkg/onchain/dns/dns.go
Trottoir: Because I use it also in the `domain.go`
Like that I have only one reference in Thyra

pkg/sc/websiteStorer.go
Trottoir: The problem is that it's not configurable : 
https://github.com/logrusorgru/textFileToGoConst/blob/c0d549f6eab77c361ef3305865e3571f8d3a1198/main.go#L152

gregLibert: Cool, I've found why this project was stopped: we should use https://pkg.go.dev/embed@master. In a new issue ?

int/api/webcreator.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func PrepareForWebsiteHandler(params operations.WebsiteCreatorPrepareParams) middleware.Responder {
-	address, err := website.PrepareForUpload(*params.Body.URL)
+	address, err := website.PrepareForUpload(*params.Body.URL, *params.Body.Nickname)
```
gregLibert, line 14: if you add to those mandatory fields the entry x-nullable: false in the swagger configuration file, go-swagger will make them as true object instead of pointer.
Maybe go-swagger goes even further by automatically handling error (this need to be check :-).

pkg/sc/websiteStorer.go
Trottoir: Issue created

https://github.com/massalabs/thyra/issues/70

pkg/node/getdatastoreentries.go
```diff
@@ -31,17 +36,38 @@ func (u JSONableSlice) MarshalJSON() ([]byte, error) {
...
+
+	response, err := DatastoreEntries(client, entries)
+	if err != nil {
+		return nil, err
+	}
+	result := *(response)
+	return &result[0], nil
+}
+
+func DatastoreEntries(client *Client, params []GetDatastoreEntriesString) (*[]DatastoreEntryResponse, error) {
```
gregLibert, line 54: In go, arrays are passed as slices. And slices are references. So you don't need to return a pointer here, the array itself is sufficiently efficient.

See https://go.dev/doc/effective_go#slices

pkg/my/domain.go
```diff
@@ -13,45 +16,51 @@ type Domain struct {
...
-	if err != nil {
-		return nil, err
 	}
 
-	dom := Domains{file: domainFile, domains: []Domain{}}
+	responses := []*models.Websites{}
+	contractAddresses, err := node.DatastoreEntries(client, params)
 
-	err = json.Unmarshal(file, &dom.domains)
+	contractAddressess := *contractAddresses
```
gregLibert, line 53: Try to avoid using additional variables that don't help understand the code.

If you solve the comment in getdatastoreentries, this variable will naturally disaper. But for your information you could have directly done : 
```go
	for i := 0; i < len(domainNames); i++ {
		response := models.Websites{
			Address: string((*contractAddresses)[i].CandidateValue),
			Name:    domainNames[i],
		}
		responses = append(responses, &response)
	}
```

gregLibert (changes requested):

PR #86, opened 2022-09-06
Add build job to CI.

thomas-senechal:


.github/workflows/CI.yml
```diff
@@ -22,3 +22,30 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         continue-on-error: true
+
+  build:
+    name: build and upload artifacts
+    strategy:
+      matrix:
```
gregLibert, line 29:
```suggestion
      target:
```

.github/workflows/CI.yml
```diff
@@ -22,3 +22,30 @@ jobs:
...
+            os: darwin
+          - arch: arm64
+            os: darwin
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.19
+      - uses: actions/checkout@v3
+      - name: Build binary for ${{ matrix.os }}
```
gregLibert, line 45:
```suggestion
      - name: Build binary for ${{ matrix.os }} on ${{ matrix.arch}}
```

gregLibert (changes requested):

.github/workflows/CI.yml
thomas-senechal: `matrix` is a keyword that (AFAIK) cannot be changed. ([Strategy Doc](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix), [Matrix Doc](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs))
I can try to make that change, and revert the change if it does not work, but I really don't think it will work.

Issue #81, opened 2022-09-06
[BACK/SC] - Change upload chunk logic


Trottoir:
**AS IS :** 

- We upload the first chunk with `initializeWebsite` in the dataEntry `massa_web` key of the `websiteCreator` contract
- Then for each following chunks, we append  in the dataEntry `massa_web` key of the `websiteCreator` contract with `appendBytesToWebsite`
- Between each chunks, we are waiting for an event to assert that the chunk uploading has ended
**TO BE :** 
- Upload each chunk with a new paramater in the smart-contract @Json struct => add an offset
- Upload each chunk with the key : ,  (...) following the chunkId
ex : 
	dataEntry "massa_web0" => chunk 0 : { data : "XXXX", offset : 0 }
	dataEntry "massa_web1" => chunk 1 : { data : "XXXX", offset : 260000 }
	dataEntry "massa_web2" => chunk 2 : { data : "XXXX", offset : 520000 }

gregLibert:
@Trottoir Don't you think that:
- you will need the total size of the data to reserve it in the smart contrat ?
You can avoid that by extending before/after the first/last element of the byte array, but I think the code will be complicated.
- this mechanism should be implemented at a storage level ?
We are working right now on website, but we could image using that for storing other data.
- to follow chuck processing we will need an event manager ?
This event manager could also expose API for third-party applications, extension event triggering mechanism to the outside world of the node.
- all the chunks could be sent in parallel with a retry mechanism (to be linked with the expiry).

Trottoir:
- We don't need the total data size as we are not registering everything in one key anymore `massa_web` => [massa_0, `massa_1 `... `massa_n`]
 This is also linked to this Issue [https://github.com/massalabs/thyra/issues/82](https://github.com/massalabs/thyra/issues/82)
- Totally, for me it's more than a website uploader, it's a "Blockchain data storage", as you explain we can register everything we want through this method and we'll be able to restitute the zip file to the user instead of rendering it as a webite for instance.
- We don't need that kind of feature to monitor that, for now we are registering a key `actual_chunk` & a `total_chunks` in the SC storage to determine the uploading status. 
**But in fact** we definetly need an event listener system that can be customizable & so on.
- Are you sure parrallel thread is usefull as we can only use the `CallFunction` that returns an operation id in all cases ? 
   I do agree about the retry mechanism

gregLibert:
[...]

PR #94, opened 2022-09-08
Add getter status on website creator

Trottoir:


int/api/webcreator.go
```diff
@@ -99,3 +99,17 @@ func UploadWebsiteHandler(params operations.WebsiteCreatorUploadParams) middlewa
 			Name:    "Name",
 			Address: params.Address})
 }
+
+func GetUploadStateHandler(params operations.WebsiteCreatorGetterStatusParams) middleware.Responder {
```
gregLibert, line 103: No Get into the getter name. See https://go.dev/doc/effective_go#Getters

pkg/front/errors.go
```diff
@@ -1,28 +1,34 @@
...
-	['Wallet-0002', 'Wrong wallet password'],
-	['Wallet-0003', 'Error while retrieving the specified wallet'],
-	['Wallet-1001', 'A wallet cannot be created without nickname'],
-	['Wallet-1002', 'A wallet cannot be created without password'],
-	['Wallet-1003', 'A technical issue occured while creating the wallet'],
-	['Wallet-2001', 'Specify the wallet wallet address to delete'],
-	['Wallet-2002', 'Error while trying to delete the wallet, may be opened by an other program'],
-	['Wallet-3001', 'Error while trying to import the wallet file'],
-	['Wallet-4001', 'Error while trying to fetch all your wallets'],
+	['Wallet-0001', 'That nickname is taken. Try Another'],
```
gregLibert, line 9: @SlnPons can you check that ?

pkg/onchain/website/uploader.go
```diff
@@ -30,7 +32,8 @@ func PrepareForUpload(url string, wallet *wallet.Wallet) (string, error) {
 }
 
 type UploadWebsiteParam struct {
-	Data string `json:"data"`
+	Data        string `json:"data"`
```
gregLibert, line 35: This feal weird : don't you need an offset to identify the chunk ?

pkg/onchain/website/uploader.go
```diff
@@ -55,6 +58,42 @@ func Upload(atAddress string, content string, wallet *wallet.Wallet) (string, er
 	return "1", nil
 }
 
+func Status(websiteCreator string) (*models.UploadState, error) {
+	client := node.NewDefaultClient()
```
gregLibert, line 62: I'm not sure it's a good idea to set the client here. When we will need to set a specific node IP, a lot of refacto will be needed...

pkg/onchain/website/uploader.go
```diff
@@ -55,6 +58,42 @@ func Upload(atAddress string, content string, wallet *wallet.Wallet) (string, er
 	return "1", nil
 }
 
+func Status(websiteCreator string) (*models.UploadState, error) {
```
gregLibert, line 61: Can we have a discussion about how you are expecting this to work functionaly ?

resource/api/swagger.yml
```diff
@@ -222,6 +222,30 @@ paths:
...
             $ref: '#/definitions/Error'
+  /websiteCreator/state/{contractAddress}:
+    get:
+      operationId: websiteCreatorGetterStatus
+      parameters:
+        - in: path
+          name: contractAddress
+          type: string
+          required: true
+          description: Website creator contract's address
```
gregLibert, line 233: @SlnPons I don't think that this descirption is sufficient for an external developer to understand this API endpoint. wdyt?

resource/api/swagger.yml
```diff
@@ -222,6 +222,30 @@ paths:
...
           schema:
             $ref: '#/definitions/Error'
+  /websiteCreator/state/{contractAddress}:
+    get:
+      operationId: websiteCreatorGetterStatus
+      parameters:
+        - in: path
+          name: contractAddress
+          type: string
+          required: true
```
gregLibert, line 232: if you are using required: true, x-nullable:false should be set i think ?

resource/html/front/errors.js
```diff
@@ -1,21 +1,27 @@
...
-	['Wallet-0002', 'Wrong wallet password'],
-	['Wallet-0003', 'Error while retrieving the specified wallet'],
-	['Wallet-1001', 'A wallet cannot be created without nickname'],
-	['Wallet-1002', 'A wallet cannot be created without password'],
-	['Wallet-1003', 'A technical issue occured while creating the wallet'],
-	['Wallet-2001', 'Specify the wallet wallet address to delete'],
-	['Wallet-2002', 'Error while trying to delete the wallet, may be opened by an other program'],
-	['Wallet-3001', 'Error while trying to import the wallet file'],
-	['Wallet-4001', 'Error while trying to fetch all your wallets'],
+	['Wallet-0001', 'That nickname is taken. Try Another'],
```
gregLibert, line 2: @SlnPons can you have a look at that plz ?

gregLibert (changes requested):

pkg/front/errors.go
SlnPons: You're right. Atm, there is not restriction about the name. But this will be implemented with TestNet 14 and so is supposed to be part of this iteration.

resource/api/swagger.yml
SlnPons: I've to say I don't understand it myself. Tbh, there is an important effort to make and time to invest on the documentation and clarity of the fct ==> should be part of DoD and technical standards.

resource/html/front/errors.js
```diff
@@ -1,21 +1,27 @@
...
-	['Domains-0001', 'Error while fetching DNS owned'],
-	['Domains-0002', 'Error while fetching addresses of DNS'],
+	['Domains-0001', 'Error while looking for your domain names'],
+	['Domains-0002', 'Error while connecting your domain and smart contract’s address'],
 
-	['WebCreator-0001', 'Creation of Webcreator Smart Contract failed'],
-	['WebCreator-1001', 'Reading of ZIP file failed'],
-	['WebCreator-1002', 'Sending of ZIP file on blockchain failed'],
+	['WebCreator-0001', 'Error while creating your website container'],
+	['WebCreator-1001', 'Impossible to read you ZIP file. Try again'],
```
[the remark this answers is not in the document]
SlnPons: Is it possible, yes. Will it change anything? probably not. Keen to discuss about error management best practices.

pkg/front/errors.go
gregLibert: Sorry, my request was unclear : can you check all the error messages :-)

resource/html/front/errors.js
gregLibert: Same here, not just this error but can you have a look at all the error messages ?

pkg/onchain/website/uploader.go
Trottoir: Not yet, it will be handled by this issue : 
[https://github.com/massalabs/thyra/issues/81](https://github.com/massalabs/thyra/issues/81)

pkg/onchain/website/uploader.go
Trottoir: To be treated here as we have this usecase in several places of the code

https://github.com/massalabs/thyra/issues/96

resource/api/swagger.yml
Trottoir: I tried and it changes nothing in generated files
I think that it is useless as it's a path parameter, so it has to be != null

gregLibert: good point. But adding it will be consistent with all the other endpoints, so if it's not harmful i prefer to add it :-).

pkg/front/errors.go
Trottoir: So as Solène already check that error messages, can we resolve that conversation ?

resource/api/swagger.yml
Trottoir: It's Updated, wdyt ?

gregLibert: @SlnPons this question is for you :-)

gregLibert (approved):

PR #104, opened 2022-09-10
lints all files

gregLibert:

PR #105, opened 2022-09-12
95 back modify websitecreatorprepare put

SebastienValla:
Need to be merged with the frontend.

AS IS :
The endpoint ONLY create the smart contract and set a resolver name on the dnsContract

TO BE :
The endpoint creates the smart contract, reserve the dnsName AND upload the website.
The /websiteCreaor/prepare POST needs to stay in order to be able to upload the content of the websiteCreator
add x-nullable : false when required : true
Add pattern on DNS name


int/api/webcreator.go
```diff
@@ -45,12 +45,35 @@ func PrepareForWebsiteHandler(params operations.WebsiteCreatorPrepareParams) mid
 				})
 	}
 
+	archive, err := ioutil.ReadAll(params.Zipfile)
```
Trottoir, line 48: Change `ioutil`  by `io` or `os` as it's described here:  
[https://pkg.go.dev/io/ioutil](https://pkg.go.dev/io/ioutil)

Trottoir (changes requested):

int/api/webcreator.go
SebastienValla: c'est fait

Trottoir (approved):

gregLibert:
Issue #104 must be close before!

resource/api/swagger.yml
```diff
@@ -208,60 +210,65 @@ paths:
...
-              url:
-                description: URL without '.', capitals letters and specifics characters
-                type: string
-                x-nullable: false
-              nickname:
-                description: Wallet's nickname to be used for preparing the website
-                type: string
-                x-nullable: false
+          x-nullable: false
+          description: URL without '.', capitals letters and specifics characters
```
gregLibert, line 239:
```suggestion
          description: URL without dot (.), upper case and special characters
```

resource/api/swagger.yml
```diff
@@ -208,60 +210,65 @@ paths:
...
       operationId: websiteCreatorPrepare
+      consumes:
+        - multipart/form-data
       parameters:
-        - in: body
-          name: body
+        - in: formData
+          name: url
+          type: string
+          pattern: "^[a-z0-9]+$"
```
gregLibert, line 236: what happens if you don't follow this rule ? Does the middle reject the request directly ?

resource/api/swagger.yml
```diff
@@ -208,60 +210,65 @@ paths:
...
           schema:
-            $ref: '#/definitions/Error'
-        '500':
+            $ref: "#/definitions/Error"
+        "500":
           description: Internal Server Error - The server has encountered a situation it does not know how to handle.
           schema:
-            $ref: '#/definitions/Error'
+            $ref: "#/definitions/Error"
   /websiteCreator/prepare:
```
gregLibert, line 227: @SlnPons can you validate this swagger endpoint ?

resource/api/swagger.yml
```diff
@@ -208,60 +210,65 @@ paths:
...
-                type: string
-                x-nullable: false
+          x-nullable: false
+          description: URL without '.', capitals letters and specifics characters
+        - in: formData
+          name: nickname
+          type: string
+          required: true
+          x-nullable: false
+          description: Wallet's nickname to be used for receiving the website
```
gregLibert, line 245:
```suggestion
          description: Name of the Wallet in which the website will be deployed.
```

resource/api/swagger.yml
```diff
@@ -9,7 +9,7 @@ produces:
 schemes:
   - http
   - https
-swagger: '2.0'
+swagger: "2.0"
```
gregLibert, line 12: Can we find a linter or a best practice for swagger file to avoid this kind of things ?

gregLibert (changes requested):

resource/api/swagger.yml
SebastienValla: you have an error 605 : url need to follow ^[a-z0-9]+$ pattern

Issue #70, opened 2022-08-31
[BACK] - Replace the file generator textFileToGoConst


Trottoir:
Replace the textFileToGoConst by a lib that is still maintained.

The reason is that on each file created, a header is created to summarize the generation information => This is messing up with Github and we want to remove it.

gregLibert:
Not possible right now because of : https://github.com/golang/go/issues/46056

PR #106, opened 2022-09-12
84 front modify html websites UI upload form

SebastienValla:
to be merged with issue 95 backend
AS IS :

    2 actions are needed from the user
        for deploying the websiteCreator
        to upload the zipped website

TO BE :

    One form composed of : (check mockups)
        the dns name input
        a file uploader button
add a pattern for DNS names
add a filter for only zip


int/api/webcreator.go
```diff
@@ -45,12 +45,35 @@ func PrepareForWebsiteHandler(params operations.WebsiteCreatorPrepareParams) mid
 				})
 	}
 
+	archive, err := ioutil.ReadAll(params.Zipfile)
```
ASAPSegfault, line 48: ioutil needs to be changed to io or os. Also the package import needs to be updated to match the new package.
Source : https://pkg.go.dev/io/ioutil

SlnPons:
UI things to address: 
1. the tooltip for the website name:
- the icon needs to be bigger. We need to see that it's a i.
- the text needs to be displayed at the top of the field currently it's displayed at the bottom of the module (see attached). If it's easier, you could remove the icon and always display the text below the text box. (like Google)
image
image
2. Buttons:
a. wallet button should be 'disable' mode (eg: https://getbootstrap.com/docs/5.2/components/buttons/#disabled-state) - for the colour I suggest either the original colour (red) or grey.
b. import website button should be responsive to the text length 
3. The module itself: 
- remove the border (to see if it looks better)
- increase padding between the 'import from' and 'upload' button (should be higher than 24px to look good normally but let's see)
4. Upload button in the table --> 'edit'
5. "Decentralised website storing" title 
- Should be "Decentralised website storage" 
- Should be left aligned (currently centered)
6. Add a title to the module: "Upload your website"
- centered in the module

pkg/front/errors.go
```diff
@@ -1,7 +1,7 @@
 // GENERATED BY textFileToGoConst
```
Trottoir, line 1: No need to push that one

Trottoir (changes requested):
You need to care about what files you are pushing
When not needed, don't put generated files if only Header is changing.
We'll not have that problem anymore when that issue will be resolved => [https://github.com/massalabs/thyra/issues/70](https://github.com/massalabs/thyra/issues/70)

Also, we don't need the png & the svg info file.

When an icon is needed, pls check the icon library from bootstrap [https://icons.getbootstrap.com/](https://icons.getbootstrap.com/)

pkg/onchain/website/browser.go
```diff
@@ -42,7 +42,11 @@ func setContentType(rsc string, writer http.ResponseWriter) {
 		writer.Header().Set("Content-Type", "text/webp")
 	case ".png":
 		writer.Header().Set("Content-Type", "image/png")
+	case ".svg":
```
Trottoir, line 45: No need that anymore as we use the icon from bootstrap

pkg/front/info.go
```diff
@@ -0,0 +1,8 @@
+// GENERATED BY textFileToGoConst
```
Trottoir, line 1: Can be removed

Trottoir (approved):

ASAPSegfault (approved):

SlnPons (approved):

PR #109, opened 2022-09-15
90 front create the stepper like component display clear messaging

Trottoir:
Stepper component creation 
Call evey second to retrieve the actual upload status => To be modified with the arrival of the eventListener

New version : 

`EventListener` on js side.

We can subscribe to an event ( following event creator & data event ).
The `EventListner` will do long polling with Thyra.
Thyra will then ask every second through the `get_sc_output_events` RPC call for the event called.
When the Event is triggered by the listener, a custom callback function is called & the long polling connection is closed


gregLibert:
> If anyone knows how to correct / remove that linter issue :'(

Done [here](62e40fc8d6f01513cfbcbc48cf7f8a5d51626319).

resource/html/front/event-manager.js
```diff
@@ -0,0 +1,16 @@
...
+	async subscribe(str, address, callback) {
+		axios({
+			url: `/thyra/events/${str}/${address}`,
+			method: 'GET',
+		})
+			.then((resp) => {
+				callback(resp);
+			})
+			.catch((e) => {
+				// TODO Implement retry mechanism here
```
Thykof, line 13: we should at least `console.error(e)`

about retry mechanism, is it what we discussed together?

resource/html/front/website.js
```diff
@@ -10,353 +10,443 @@ getWebsiteDeployerSC();
...
+		document.getElementById('website-upload-refuse').style.display = 'flex';
+	} else {
+		uploadable = true;
+		document.getElementsByClassName('dns-error')[0].style.display = 'none';
+		document.getElementById('website-upload').style.display = 'flex';
+		document.getElementById('website-upload-refuse').style.display = 'none';
+	}
 });
 
 function deployWebsiteAndUpload(password) {
```
Thykof, line 302: duplicated code of above `function uploadWebsite`

int/api/events.go
```diff
@@ -0,0 +1,67 @@
...
+		return operations.NewThyraEventsGetterInternalServerError().
+			WithPayload(
+				&models.Error{
+					Code:    errorCodeGetDomainNames,
+					Message: err.Error(),
+				})
+	}
+
+	slotStart := node.Slot{
+		Period: status.LastSlot.Period - periodsBefore,
```
ASAPSegfault, line 30: Can status.LastSlot.Period be smaller than 2 ?
If that is the case we may end up with Period being negative in some cases. I don't know how it will impact the rest of the code (maybe node.Events will just throw an error) but I'm not sure this is the desired behavior.

SebastienValla (approved):

PR #111, opened 2022-09-19
Add embed package to package resource on build

Trottoir:
Modify project structure
go generate is now at root project


Thykof:
because this PR moves file and changes the structure, it might be a good idea to format all file with prettier including html css and js files. what do you think?

Trottoir:
> because this PR moves file and changes the structure, it might be a good idea to format all file with prettier including html css and js files.
> 
> also add .prettierrc.yml config file
> 
> what do you think?

Yeah I can add it

Thykof (approved):

SebastienValla (approved):

Thykof (approved):

qdrn:
Looks much clearer than before. Nice work!

I would probably put usage and install procedure on top of README as that's what 99% of the users will want. I would also put how to change the server a bit more to the top as it's also something that a lot of devs (at least) will want to use.

PR #139, opened 2022-09-26
100 add zip file filter on backend

SebastienValla:
As Is : Back allows to upload other files than zip files
To be : Do not allow to upload other files than zip on backend


SebastienValla:
Any idea how to solve the linter issue ? I did not change this line and I don't know how to solve it

int/api/webcreator.go
```diff
@@ -102,6 +115,18 @@ func UploadWebsiteHandler(params operations.WebsiteCreatorUploadParams) middlewa
 			})
 	}
 
+	contentType := http.DetectContentType(archive)
+
+	if contentType != "application/zip" {
```
Trottoir, line 120: As this part of code is the same as the one between `l 46-57`
It may be a good idea to separate it in an other function ?

ASAPSegfault (approved):
Apart from Maxime's comment stating that we can separate zip file checking logic in another function, which I agree with, it's all good to me.

int/api/webcreator.go
SebastienValla: I agree, it is done

Trottoir (approved):

PR #141, opened 2022-09-26
132 [front] limit website upload

SebastienValla:
A limit of 1.5MB should be applied onto the uploaded website.

    Display error message if file is too big
    Disable upload button if file is too big


Thykof (approved):
not so easy to read with the code formatting, it would be easier if there were 2 commits: one for changes and one for code formating; but it is ok for me.

ASAPSegfault (approved):
We must do something about code formatting but otherwise I don't see anything to change in the logic

Thykof:
@SebastienValla what are the rules you used for prettier to format the code?

gregLibert:
@SebastienValla can you solve the conflicts ?

SebastienValla:
> 
Done

PR #152, opened 2022-09-28
128 Registry "ALL domains" endpoint implementation

Trottoir:
Correction of Addresses function from ledger
Implementation of a key filter on a specified contract Package of API call related to websites & DNS ( api => websites)


int/api/errors.go
```diff
@@ -1,6 +1,7 @@
 package api
 
 const (
+	errorCodeEventListenter        = "EventLisntener-0001"
```
ASAPSegfault, line 4: Should be EventListener

ASAPSegfault (changes requested):

int/api/errors.go
Trottoir:
```suggestion
	errorCodeEventListener        = "EventLisntener-0001"
```

SebastienValla (approved):

pkg/node/getdatastoreentries.go
```diff
@@ -51,6 +51,26 @@ func DatastoreEntry(client *Client, address string, key string) (*DatastoreEntry
 	return &response[0], nil
 }
 
+// TODO need help to find name for this, this function prevent us to build the DatastoreEntriesKeysAsString if we are calling the same contract
+func DatastoreEntriesOnSameContract(client *Client, address string, keys []string) ([]DatastoreEntryResponse, error) {
```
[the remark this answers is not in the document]
Trottoir: Go for it thanks !

Thykof (approved):

Thykof (approved):

PR #158, opened 2022-10-03
136 - Add GUI for password prompt

Thykof:
# What is added?
- [x] use fyne GUI library to prompt the password for signing a transaction
- [x] center the popup
- [x] focus the popup
- [x] bug: can't close thyra
- [x] build for macos with signature
- [ ] build for macos with notarize

Screenshot 2022-10-05 at 10 14 40

# Known issues
- in thyra app, step 1 is active (with ...) while the prompt is open (https://github.com/massalabs/thyra/issues/192)
- error handling when the user click on the Cancel button in the GUI is bad (https://github.com/massalabs/thyra/issues/193)


Thykof:
@gregLibert I have squashed the commit you can merge as is if it's ok for you

Trottoir (approved):

int/api/server.go
```diff
@@ -0,0 +1,111 @@
...
+
+	localAPI.ThyraWalletHandler = operations.ThyraWalletHandlerFunc(ThyraWalletHandler)
+	localAPI.ThyraWebsiteCreatorHandler = operations.ThyraWebsiteCreatorHandlerFunc(ThyraWebsiteCreatorHandler)
+
+	server.ConfigureAPI()
+
+	if err := server.Serve(); err != nil {
+		//nolint:gocritic
+		log.Fatalln(err)
+	}
```
gregLibert, line 110: To solve the interrupt signal is not stopping thyra:
```suggestion
	}
	(*app).Quit()
```

.github/workflows/CI.yml
```diff
@@ -42,28 +48,74 @@ jobs:
           files: coverage.coverprofile
 
   build:
+    # if: ${{ github.ref == 'refs/heads/main' }}
```
thomas-senechal, line 51: Is that commented line still useful ?

gregLibert (changes requested):

thomas-senechal (changes requested):

.github/workflows/CI.yml
```diff
@@ -42,28 +48,74 @@ jobs:
...
     name: build and upload artifacts
+    env:
+      BUILD_CERTIFICATE_BASE64: ${{ secrets.BUILD_CERTIFICATE_BASE64 }}
+      P12_PASSWORD: ${{ secrets.P12_PASSWORD }}
+      BUILD_PROVISION_PROFILE_BASE64: ${{ secrets.BUILD_PROVISION_PROFILE_BASE64 }}
+      KEYCHAIN_PASSWORD: ${{ secrets.KEYCHAIN_PASSWORD }}
+      AC_USERNAME: ${{ secrets.AC_USERNAME }}
```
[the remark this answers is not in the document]
Thykof: there are useful for signing and notarize, but for now we only sign. I suggest to keep like this and see in the next PR.

.github/workflows/CI.yml
Thykof: absolutely not, good catch

int/api/html/front/website.js
```diff
@@ -8,10 +8,15 @@ let uploadable = false;
 getWallets();
 getWebsiteDeployerSC();
 initializeDefaultWallet();
-setupModal();
 
 const eventManager = new EventManager();
 
+async function onSubmitDeploy(txType = 'deployWebsiteAndUpload') {
+	setTxType(txType);
+	console.log("16");
```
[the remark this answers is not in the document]
Thykof: absolutely not, good catch

thomas-senechal:
@Thykof As explained in [that comment](https://github.com/massalabs/thyra/pull/158#discussion_r992110248), I think there is an issue because most of the code you push is using tabulations instead of 4 spaces. Installating Prettier in your IDE should fix the problem.

PR #164, opened 2022-10-03
137 back send operation endpoint creation

thomas-senechal:


SebastienValla (approved):

Thykof (approved):

Trottoir (approved):

int/api/cmd/execute_function.go
```diff
@@ -13,40 +11,37 @@ import (
 	"github.com/massalabs/thyra/pkg/wallet"
 )
 
-func NewExecuteFunction(walletStorage *sync.Map) *FunctionExecuter {
-	return &FunctionExecuter{walletStorage: walletStorage}
+func NewExecuteFunction() *FunctionExecuter {
```
gregLibert, line 14: Useless now. This constructor should be remove to reduce code complexity.

int/api/cmd/execute_function.go
```diff
@@ -13,40 +11,37 @@ import (
...
+	return &FunctionExecuter{}
 }
 
-type FunctionExecuter struct {
-	walletStorage *sync.Map
-}
+type FunctionExecuter struct{}
 
 //nolint:nolintlint,ireturn
 func (f *FunctionExecuter) Handle(params operations.CmdExecuteFunctionParams) middleware.Responder {
```
gregLibert, line 21: Without the constructor, should be something like:

```suggestion
func ExecuteFunctionHandler(params operations.CmdExecuteFunctionParams) middleware.Responder 
```

int/api/cmd/execute_function.go
```diff
@@ -13,40 +11,37 @@ import (
...
 	addr, err := base58.CheckDecode((*params.Body.At)[1:])
 	if err != nil {
-		panic(err)
+		return operations.NewCmdExecuteFunctionUnprocessableEntity().WithPayload(
+			&models.Error{Code: errorCodeUnknownKeyID, Message: "Error: cannot decode Smart contract address : " + err.Error()})
```
gregLibert, line 25:
```suggestion
			&models.Error{Code: errorCodeUnknownKeyID, Message: "Error: cannot decode Smart contract address: " + err.Error()})
```

api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -396,6 +398,9 @@ paths:
               fee: 0
               keyId: default
             properties:
+              nickname:
```
gregLibert, line 401: If the field is mandatory, the attribute x-nullable: false shall be used.

int/api/cmd/execute_function.go
```diff
@@ -13,40 +11,37 @@ import (
...
 	}
 
 	addr = addr[1:]
 
-	value, exists := f.walletStorage.Load(*params.Body.Name)
-	if !exists {
+	wallet, err := wallet.Load(*params.Body.Nickname)
```
gregLibert, line 30: params.Body.Nickname is a pointer because the field Nickname is not set as nullable:false and therefore the framework consider this fields as nullable.
This field should be mandatory, don't you think ?

int/api/cmd/execute_function.go
```diff
@@ -13,40 +11,37 @@ import (
...
+	if err != nil {
 		return operations.NewCmdExecuteFunctionUnprocessableEntity().WithPayload(
 			&models.Error{
 				Code:    errorCodeUnknownKeyID,
-				Message: "Error: unknown wallet nickname : " + *params.Body.Name,
+				Message: "Error: unknown wallet nickname : " + *params.Body.Nickname,
```
gregLibert, line 35: How do you know the wallet is unknown ?
wallet.Load returns two kind of error:
- reading file error
- parsing file error

I think the best way to know the error is to return the error :-)

gregLibert (changes requested):

gregLibert (approved):

PR #189, opened 2022-10-10
180 windows installation script

ASAPSegfault:
This script download and install Thyra and Acrylic DNS Proxy on Windows.
It also configures Windows to use local Acrylic DNS
If Acrylic is already installed, the script will simply add some configuration to the Hosts file.


thyra-installer.bat
```diff
@@ -0,0 +1,60 @@
...
+    )
+    MOVE %STARTING_WORKING_DIR%Acrylic-Portable.zip %ACRYLIC_PATH%\
+    tar -xf  %ACRYLIC_PATH%\Acrylic-Portable.zip -C %ACRYLIC_PATH%
+    @REM Ideally we should check here that Acrylic has been successfully unarchived.
+    CD %ACRYLIC_PATH%
+
+    CALL "InstallAcrylicService.bat"
+
+    @REM Allows for variable value assertion inside for loops at runtime
+    @REM This for loop iterate over all connected network adapters and set their DNS.
```
Trottoir, line 37: Does that mean that I'll have the DNS setup for my Ethernet & Wifi ( for instance ) ?

ASAPSegfault: If both Wi-Fi and Ethernet network adapters are connected at the moment you execute the script, then yes it will setup the DNS for both. If only one of them is connected DNS will be setup for this one only.

Thykof (approved):

thyra-installer.bat
gregLibert: Why not configuring all network adapters ?

ASAPSegfault: If I have several network adapters and I don't want all of them to be using Acrylic DNS I should have the possibility to avoid this.
I don't really see why we should force this on the user because IMHO it's not our choice to make.
Although this choice could take the shape of a flag given to the script (Only connected net adapters / All adapters).

SebastienValla (approved):

thyra-installer.bat
gregLibert: Don't you think that it can be disturbing to have thyra working on ethernet, but not in wifi ?

ASAPSegfault: It will be for sure if people are not aware of how it works but it could be stated as a disclaimer in the installation documentation process.
On another hand we could configure all of them and the disclaimer could state that ALL network adapters will be configured to work with Acrylic DNS, connected or not.
The thing I am afraid of here is that it works great for madame Michu but may break things for more advanced users who have different DNS for different adapters (even though I don't really see the use case for this).

PR #205, opened 2022-10-17
202 create a parameter to change the limit upload size of thyra web on chain

Thykof:
usage: `UPLOAD_MAX_SIZE=3500000 go run ./cmd/thyra-server/main.go`

### DIFF
--- a/int/api/websites/webcreator.go
+++ b/int/api/websites/webcreator.go
@@ -18,6 +18,8 @@ import (
 
 const UploadMaxSize = "UPLOAD_MAX_SIZE"
 
+const defaultMaxArchiveSize = 15000
+
 func CreatePrepareForWebsiteHandler(
 	app *fyne.App,
 ) func(params operations.WebsiteCreatorPrepareParams) middleware.Responder {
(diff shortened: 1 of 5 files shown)


int/api/websites/webcreator.go
```diff
@@ -71,6 +77,21 @@ func prepareForWebsiteHandler(params operations.WebsiteCreatorPrepareParams, app
...
+func GetMaxArchiveSize() int {
+	strVar := os.Getenv(UploadMaxSize)
+
+	if strVar == "" {
+		return defaultMaxArchiveSize
+	}
+
+	intVar, err := strconv.Atoi(strVar)
+	if err != nil {
+		panic(err)
```
Trottoir, line 89: As it's called from an HTTP call, we need to return an error in the

Trottoir (changes requested):

int/api/websites/webcreator.go
```diff
@@ -71,6 +77,21 @@ func prepareForWebsiteHandler(params operations.WebsiteCreatorPrepareParams, app
 			})
 }
 
+func GetMaxArchiveSize() int {
+	strVar := os.Getenv(UploadMaxSize)
```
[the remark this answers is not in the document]
Thykof: you'r right

int/api/websites/webcreator.go
Thykof: I a pushed a commit to return the default value

Trottoir (approved):

SebastienValla (approved):

Trottoir (approved):

PR #213, opened 2022-10-19
207 create installation script for linux

thomas-senechal:


scripts/linux_install.sh
```diff
@@ -0,0 +1,77 @@
...
+        *)      fatal "$(uname -m) unsupported." ;;
+    esac
+}
+
+check_supported_distro () {
+    grep '^ID.*=.*ubuntu' /etc/os-release &> /dev/null || fatal "Aborting: You are using an unsupported distribution."
+}
+
+install_thyra () {
+    arch="$(architecture_version)" || exit 1
```
Thykof, line 25: could it exit without error message? should we do here `fatal` like at other places?

thomas-senechal: There is no need for an error message here since it's already printed [here](https://github.com/massalabs/thyra/blob/0f345d56268b5f7eace23fe8c08a3b2de9b4f801/scripts/linux_install.sh#L16).

Thykof: Okey then I think that `|| exit 1` is not necessary

thomas-senechal: If we remove the `|| exit 1`, the script won't stop at that line and will continue to run the next steps of the function. Those `exit 1` are like `return`.

Thykof: I said that because there is an `exit 1` in the `fatal` function

Thykof:
I have tested on macos, I got the espected error about the arch.
I have tested on linux ubuntu 22, it works

Thykof (approved):

scripts/linux_install.sh
```diff
@@ -0,0 +1,77 @@
...
+architecture_version () {
+    case $(uname -m) in
+        x86_64) echo "amd64" ;;
+        # arm64)  echo "arm64" ;; ARM64 is unsupported on Linux for now
+        *)      fatal "$(uname -m) unsupported." ;;
+    esac
+}
+
+check_supported_distro () {
+    grep '^ID.*=.*ubuntu' /etc/os-release &> /dev/null || fatal "Aborting: You are using an unsupported distribution."
```
peterjah, line 21: any reason to support only Ubuntu ?? maybe rely on ID_LIKE key (from /etc/os-release)  will be better? to accept all Debian like distros.

scripts/linux_install.sh
```diff
@@ -0,0 +1,77 @@
...
+}
+
+install_thyra () {
+    arch="$(architecture_version)" || exit 1
+    curl -s -L "${BINARY_URL}_${arch}" -o thyra-server || fatal "binary download failed."
+
+    chmod +x thyra-server || fatal "change to executable failed."
+    (ls /usr/local/bin/ &> /dev/null || sudo mkdir -p /usr/local/bin) || fatal "/usr/local/bin creation failed."
```
peterjah, line 29: i think ```ls /usr/local/bin/ &> /dev/null``` is useless here

peterjah: and also i doubt there is a chance that /usr/local/bin does not exists. thats mean the whole line is useless for me

peterjah (changes requested):
great work :)

scripts/linux_install.sh
peterjah: image

peterjah: fyi a screenshot of the os-release file from my vps running on debian.
image

thomas-senechal: >any reason to support only Ubuntu ?? 

It has been decided to only support Ubuntu for now since it's one of the most used and known distribution. We are planning to had more supported distros following user demands. 

> maybe rely on ID_LIKE key (from /etc/os-release) will be better? to accept all Debian like distros
In fact, I am relying on `ID` and `ID_LIKE`. My regex `'^ID.*=.*ubuntu'` checks for ubuntu in both `ID` and `ID_LIKE`. Which allows us to support Ubuntu based distributions such as PopOS, Linux Mint, Zorin OS...

scripts/linux_install.sh
thomas-senechal: Done here https://github.com/massalabs/thyra/pull/213/commits/2e613911c5bc7d47423bbf1c16fda6f84f4ff505

scripts/linux_install.sh
peterjah: fair enough :)

peterjah (approved):

PR #216, opened 2022-10-19
add air and update readme

Thykof:


ASAPSegfault (commented):
Here [readme](https://github.com/cosmtrek/air) it is stated that .air.conf will be deprecated soon.
Maybe we should go with .air.toml as adviced right away ?

Thykof:
Good catch @ASAPSegfault it have renamed the configuration file

ASAPSegfault (approved):

Trottoir (approved):
Lovely <3

PR #229, opened 2022-10-24
Updating Installation MD

0xMazout:
Bat Install is the main thing to install
Atm process is not working
- Adding line to prefer using bat than process


0xMazout:
Was send with the #230 i close the pr

PR #230, opened 2022-10-25
Add errors empty pass word cancel gui

0xMazout:
Adding two new error about the gui window 

Password is empty 
Canceled action


Trottoir (approved):

Trottoir:
@0xMazout you need to correct the linter here.

You can check in `Files changed` in order to see what has been triggered by it !

pkg/wallet/wallet.go
```diff
@@ -80,6 +80,10 @@ func (w *Wallet) Unprotect(password string, keyPairIndex uint8) error {
 	if err != nil {
 		return fmt.Errorf("intializing block ciphering: %w", err)
 	}
+	// length control on password
+	if len(password) == 0 {
+		return fmt.Errorf("password is empty. try again %w", errors.New("password is empty. try again"))
+	}
```
Thykof, line 86: length control on password is done 2 times, we should not do it here but rather before calling Unprotect

Thykof (commented):
I see the HTTP response is OK but I don't see the error messages in the front end

0xMazout:
For me right now it's okay and it's working maybe we can move it to "Code review"

Thykof:
It works, good job!

Thykof (approved):

int/api/cmd/execute_function.go
```diff
@@ -38,9 +39,17 @@ func ExecuteFunctionHandler(params operations.CmdExecuteFunctionParams, app *fyn
...
-	password := gui.AskPassword(wallet.Nickname, app)
+	password, status := gui.AskPassword(wallet.Nickname, app)
+	if !status {
+		return createInternalServerError(websites.ErrorCodeWalletCanceledAction, websites.ErrorCodeWalletCanceledAction)
+	}
 
 	err = wallet.Unprotect(password, 0)
+
+	if len(password) == 0 {
+		return createInternalServerError(websites.ErrorCodeWalletPasswordEmpty, err.Error())
```
gregLibert, line 50: Are you sure there is an error here ?

int/api/websites/webcreator.go
```diff
@@ -35,9 +35,17 @@ func prepareForWebsiteHandler(params operations.WebsiteCreatorPrepareParams, app
...
 
-	password := gui.AskPassword(wallet.Nickname, app)
+	password, status := gui.AskPassword(wallet.Nickname, app)
+	if !status {
+		return createInternalServerError(ErrorCodeWalletCanceledAction, ErrorCodeWalletCanceledAction)
+	}
 
 	err = wallet.Unprotect(password, 0)
+
+	if len(password) == 0 {
```
gregLibert, line 45: Idem

pkg/gui/password.go
```diff
@@ -18,8 +18,9 @@ func AskPasswordDeleteWallet(nickname string, app *fyne.App) string {
 // inspired by https://hackernoon.com/asyncawait-in-golang-an-introductory-guide-ol1e34sg
 
 // Thyra password input dialog.
-func PasswordDialog(nickname string, app *fyne.App) chan string {
+func PasswordDialog(nickname string, app *fyne.App) (chan string, chan bool) {
```
gregLibert, line 21: Why did you prefer to use a boom instead of a string pointer if the string is optional ?

gregLibert (changes requested):

int/api/cmd/execute_function.go
0xMazout: I'm not sure about the question, but the target was to upgrade the error wich was the same as "wrong password" to have a an error specialy when the password is Empty. And then when password is empty it display an "Empty password" error. 

If it's needed to make some adjustments let me know

pkg/gui/password.go
0xMazout: Because cancel and empty password produce the same result. 

which is passwordText is an empty string and to catch the difference between the two action, i had to create a "trigger" or a "flag" to catch it on the other side. Then i used adding a bool. 

But if you think there is a better way with less impact. Let's change it

int/api/cmd/execute_function.go
```diff
@@ -39,8 +40,18 @@ func ExecuteFunctionHandler(params operations.CmdExecuteFunctionParams, app *fyn
 	}
 
 	password := gui.AskPassword(wallet.Nickname, app)
```
gregLibert, line 42: In go you can returned multiples values. The way to handle error in go is very strict : the last returned value should be an error.

This line should be something like:
```suggestion
	clearPassword, err := gui.AskPassword(wallet.Nickname, app)
```

pkg/gui/password.go
```diff
@@ -36,10 +44,10 @@ func PasswordDialog(nickname string, app *fyne.App) chan string {
 		Items: items,
 		OnSubmit: func() {
 			window.Hide()
-			passwordText <- password.Text
+			passwordEntry <- PasswordEntry{password.Text, nil}
 		},
 		OnCancel: func() {
-			passwordText <- ""
+			passwordEntry <- PasswordEntry{password.Text, errors.New("canceled by user")}
```
gregLibert, line 50: A couple of things here:
- when you use composite litteral, I personnaly find it easier to use labels. That way I don't have to get the struct in mind. More information on composite litterals [here](https://go.dev/doc/effective_go#composite_literals).
- using password.Text instead of an empty string makes me wonder what is the default value of password.Text. And i don't see it in the screen.

Mu suggestion is:
```suggestion
			passwordEntry <- PasswordEntry{"", errors.New("password entry cancelled by the user")}
```

pkg/gui/password.go
```diff
@@ -99,8 +107,10 @@ func PasswordDeleteDialog(nickname string, app *fyne.App) chan string {
 }
 
 // This function is blocking, it returns when the user submit or cancel the form.
-func Password(nickname string, app *fyne.App) string {
-	return <-PasswordDialog(nickname, app)
+func Password(nickname string, app *fyne.App) PasswordEntry {
```
gregLibert, line 110: Your password function is exposing its internal logic: You should return a string and an error.

As explained above the code consumming this function feel weird because of this.

My suggestion: 
```suggestion
func Password(nickname string, app *fyne.App) (string, error) {
```

gregLibert (changes requested):

SebastienValla (approved):

gregLibert (approved):

PR #232, opened 2022-10-26
226 error management improvements

thomas-senechal:


Thykof (approved):
good job, but I got this error on submitting the form without wallet selected:
```
Uncaught (in promise) TypeError: getWallet(...) is undefined
    initStepper http://my.massa/thyra/websiteCreator/website.js:376
```

thomas-senechal:
> good job, but I got this error on submitting the form without wallet selected:

Thank you for your review @Thykof . While fixing this error, I noticed more issues that should be handled. So I will put that PR back In Progress to fix those bugs.

Thykof (approved):

SebastienValla (approved):

int/api/html/front/event-manager.js
```diff
@@ -11,7 +11,7 @@ class EventManager {
             })
             .catch((e) => {
                 // TODO Implement retry mechanism here
-                console.error(e);
+                console.error(e.response.data);
```
gregLibert, line 14: And what happens if the error is not in a server response?
You will get an exception 'e.response' is not defined...
You can find in the code example [here](https://axios-http.com/docs/handling_errors) different keys to get according to the different scenarios.
I don't know if it's exhaustive though...

int/api/html/front/wallet.js
```diff
@@ -30,6 +30,7 @@ async function importWallet(wallet) {
             wallets.push(wallet);
         })
         .catch((e) => {
+            console.error(e.response.data);
```
gregLibert, line 33: idem here and below :-)

gregLibert (changes requested):

int/api/html/front/errors.js
```diff
@@ -27,3 +29,24 @@ const errorCodes = new Map([
...
+    } else if (error.request) {
+        // The request was made but no response was received
+        console.error(error.request);
+    } else if (error.message) {
+        // Something happened in setting up the request that triggered an Error
+        console.error(error.message);
+    } else {
+        console.error(error);
+    }
+}
```
gregLibert, line 52:
```suggestion
// If the error is from Thyra, we display the error to the user and log the details in the console.
// Otherwise, we simply display the details in the console.
function handleAjaxError(error) {

    if (error.response && error.response.data) {
            if (error.response.data.code) {
                errorAlert(getErrorMessage(error.response.data.code));
            }
            console.error("Thyra error:", error.response.data);
    } else {
        console.error(error);
    }
}
```

gregLibert: @SlnPons I'm not sure it's a good idea to don't share with the end user when something went wrong even if the error is not from thyra.

Wdyt?

gregLibert (changes requested):

int/api/html/front/errors.js
SlnPons: @gregLibert agreed. All errors should be displayed to the end user so that they understand what's going on. 
Just for my understanding currently the errors can come from:
* Thyra 
* Node API (the blockchain)
Do I miss anything?

gregLibert: The error can come:
- from the library we use to send request to thyra
- from the browser itself
- from the network
- from the framework we use on thyra side
- from our code on thyra

Only errors from the last case are screened out right now with the proposed code.

SlnPons: Ok, thank you. I don't know the best practices about how to manage and list all these errors (same library, separate ones,...). But I know we need to have them listed somewhere AND handled beautifully for us and for the user.

How do you think we should handle them? 
Keeping in mind that the different objects (ie: wallet, storage...) will be removed from Thyra to act as external modules and be used in plug-in.

gregLibert (approved):

PR #242, opened 2022-11-03
Adds Web on-chain diagram

SlnPons:


doc/Wallet.plantuml
```diff
@@ -0,0 +1,94 @@
+@startuml
+participant "End user" as user
+participant "Browser" as ui
+participant "Thyra server" as thyra
+participant "Wallet" as w
+participant "Node (API)" as node
+
```
gregLibert, line 7:
```suggestion
title "Web on chain sequence diagram"
```

doc/WebOnChain.plantuml
```diff
@@ -0,0 +1,186 @@
...
+participant "End user" as user
+participant "Browser" as ui
+participant "Thyra server" as thyra
+participant "Hosting websites" as web
+participant "Wallet" as w
+participant "DNS" as dns
+participant "Event Manager" as em
+participant "Storage" as storage
+participant "Node (API)" as node
+
```
gregLibert, line 11:
```suggestion
title "= **Website hosting on blockchain**"
```

doc/WebOnChain.plantuml
```diff
@@ -0,0 +1,186 @@
...
+user -> ui: click on install 'Web on-chain' plug-in
```
gregLibert, line 12: You should add a [divisor](https://plantuml.com/fr/sequence-diagram#d4b2df53a72661cc) here to make thing clearer.

Furthermore you should move this part to a dedicated diagram. This schema is to big.
```suggestion
== Plug-in installation ==

user -> ui: click on install 'Web on-chain' plug-in
```

doc/WebOnChain.plantuml
```diff
@@ -0,0 +1,186 @@
...
+ui -> thyra: GET /plugin/webOnChain
+thyra -> web: GET /plugin/webOnChain
```
gregLibert, line 14: So you need to have the webOnChain plug-in to be installed to install the webOnChain plug-in ?

doc/WebOnChain.plantuml
```diff
@@ -0,0 +1,186 @@
...
+web -> thyra: OK response
+thyra -> ui: display landing page WebOnChain.massa
```
gregLibert, line 16: How does it work ?
- When you create the webOnChain plugin, you natively have the webonchain.massa domain ?
- You need to by it manually ?
- Should we reserved/block `plugin` resource at the DNS level ?

Final note: URL are case sensitive. Do we plan to allow webOnChain, WebOnChain, webonChain, webonchain to be valid values and all pointing to different urls ?

doc/WebOnChain.plantuml
```diff
@@ -0,0 +1,186 @@
...
+ui -> user: display landing page WebOnChain.massa
+
+user -> ui: Click on "websites" page
+ui -> thyra: GET /plugin/wallet
+thyra -> web: GET /plugin/wallet
+web -> thyra: GET /plugin/wallet/{status}
+thyra -> w: GET /plugin/wallet/{status}
+w -> w: is wallet status active?
```
gregLibert, line 24: What appens if there is more than one wallets ? no wallet ?

doc/WebOnChain.plantuml
```diff
@@ -0,0 +1,186 @@
...
+w -> thyra: KO response - inactive status
+thyra -> web: KO response - inactive status
+web -> ui: display KO response - inactive status
+ui -> user: Error "Connect to your wallet" displayed
+w -> thyra: OK response
```
gregLibert, line 29:
```suggestion
alt wallet is inactive
    w -> thyra: KO response - inactive status
    thyra -> web: KO response - inactive status
    web -> ui: display KO response - inactive status
    ui -> user: Error "Connect to your wallet" displayed
else wallet is active
    w -> thyra: OK response
end
```

doc/WebOnChain.plantuml
SlnPons: Hmm, no. I wanted to show that somehow when the user click on install (so from ui -> thyra) the name, port or sth else must be clearly identified and known as a reference to the web on-chain plug-in so that the installation starts.

doc/WebOnChain.plantuml
SlnPons: ack, will share an updated and lighter version.

doc/WebOnChain.plantuml
SlnPons: I think I mistaken the installation process....to install the plug-in we need to download the packages (installation script, no sure about the best vocabulary here). So the steps here aren't relevant. 
Will update the diagrams with a suggestion. Thanks :)

doc/WebOnChain.plantuml
SlnPons: ack, will suggest something to handle that.

PR #246, opened 2022-11-07
improve quit gui

Thykof:
The application was not quit properly. This PR defers the call to the quit function to execute it in any case.

Before this PR, the bug was that the application was still running event after submitting the password.
Screenshot 2022-11-07 at 10 53 25


thomas-senechal (approved):
Working well on MacOS.

PR #248, opened 2022-11-07
fix redirection for custom port use

peterjah:


thomas-senechal (approved):
Everything seems to work as expected now. Nice work !

SebastienValla (approved):

PR #273, opened 2022-11-17
262 add a version option at launch

sydhds:
Edit @thomas-senechal : 
I remade the implementation following [@gregLibert comment](https://github.com/massalabs/thyra/pull/273#pullrequestreview-1189535341). 
I also remade the flags handling following a meeting with @gregLibert .


thomas-senechal (approved):
Very good work !

.github/workflows/CI.yml
```diff
@@ -91,6 +91,9 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
       - uses: actions/checkout@v3
+      - name: Update version
+        run: |
+          echo -e "package api\n\nconst version_str=\"${GITHUB_REF_NAME}\"" > int/api/server_version.go
```
Thykof, line 96: should we do this also in CD workflow?

Thykof (approved):

.github/workflows/CI.yml
thomas-senechal: You are right. I missed that. I thought we were using the CI workflow in the CD one to build the binaries. But we don't.

gregLibert (changes requested):
:no_good: 
I'm not sure we are using the right method to do this:
- we are writing go code from the shell when we could have used:
  * an ldflags as explained [here](https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications), or
  * an embedded file as explained [here](https://levelup.gitconnected.com/a-better-way-than-ldflags-to-add-a-build-version-to-your-go-binaries-2258ce419d2d)
- we lose readability by moving all other server commands to an else branch
- We duplicate code such as `defer (*app).Quit()`.

My feeling is that we are adding legacy code here when we should start cleaning things up by following a boyscout approach...

As we are not in the hackathon rush anymore, can someone from the innoteam can take it over?

int/api/server.go
```diff
@@ -18,31 +17,28 @@ import (
...
+	Port              int
+	TLSPort           int
+	TLSCertificate    string
+	TLSCertificateKey string
+	MassaNodeServer   string
+	Version           bool
+}
 
-	flag.Parse()
+func parseFlags(server *restapi.Server, startFlags StartServerFlags) {
```
Thykof, line 29: I think it's confusing to have 2 parseFlags function, we should name them differently

Thykof (changes requested):

int/api/server.go
peterjah: i guess it is more something about configuring the server rather than parsing flag

int/api/server.go
```diff
@@ -60,10 +56,23 @@ func parseNetworkFlag(massaNodeServerPtr *string) {
 	os.Setenv("MASSA_NODE_URL", *massaNodeServerPtr)
 }
 
+func stopServer(app *fyne.App, server *restapi.Server, manager *pluginmanager.PluginManager) {
+	defer func() {
```
peterjah, line 60: I think the "defer" is useless here.. just stops the server directly

peterjah: defer is used to set a callback when the function returns. here your fuction does not do anything else

int/api/server.go
```diff
@@ -103,29 +110,20 @@ func StartServer(app *fyne.App) {
...
-	parseFlags(server)
+	setAPIFlags(server, startFlags)
 
 	// Run plugins
 	manager, err := pluginmanager.New(server.Port, server.TLSPort)
 	if err != nil {
 		log.Fatalln(err)
 	}
 
-	defer func() {
```
Thykof, line 114: I think here we should do 

```
defer stopServer(app, server, manager)
```

int/api/server.go
Thykof: I disagree, see my comment, defer is useful to stop the server if the code panic while handling an error

thomas-senechal: Oh okay. Then I will do as you said in your comment. Thanks a lot !

Thykof (approved):

Thykof (approved):
🤷 LGTM

peterjah (approved):

gregLibert (approved):
:see_no_evil:

PR #274, opened 2022-11-17
263-display-the-selected-node-server-value-at-launch

sydhds:


int/api/server.go
```diff
@@ -76,9 +79,19 @@ func StartServer(app *fyne.App) {
...
 	parseFlags(server)
 
+	// Display info about node server
+	client := node.NewDefaultClient()
+	status, err := node.Status(client)
+
+	if err == nil {
+		//nolint:forbidigo
+		fmt.Println("Running on", *status.Version)
+	}
```
thomas-senechal, line 89: I think we should display the URL of the used node since the issue was about displaying the `node-server` option.

```suggestion
	nodeVersion := "unknown"
	if err == nil {
		nodeVersion = *status.Version
	} else {
		log.Println("Could not get node version:", err)
	}

	log.Printf("Connected to node server %s (version %s)\n", os.Getenv("MASSA_NODE_URL"), nodeVersion)
```

thomas-senechal: Note that if you apply the suggested changes, you might have a linter error since there is no more need for the `fmt` import.

thomas-senechal (changes requested):
Great work !

Trottoir (approved):

thomas-senechal (approved):

PR #277, opened 2022-11-21
add check if .html file exist in source of zip file, otherwise return Error

B-Naoufal:
… an error WebCreator-1005

### DIFF
--- a/int/api/websites/webcreator.go
+++ b/int/api/websites/webcreator.go
@@ -86,7 +86,7 @@ func prepareForWebsiteHandler(params operations.WebsiteCreatorPrepareParams, app
 	FilesOfArchive := listArchiveContent(zipReader)
 
 	if !contains(FilesOfArchive, "index.html") {
-		return createInternalServerError(errorCodeWebCreatorHTMLNotInSource, err.Error())
+		return createInternalServerError(errorCodeWebCreatorHTMLNotInSource, errorCodeWebCreatorHTMLNotInSource)
 	}
 
 	b64 := base64.StdEncoding.EncodeToString(archive)
--- a/int/api/websites/webcreator.go
+++ b/int/api/websites/webcreator.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+
 	"golang.org/x/exp/slices"
 
 	"fyne.io/fyne/v2"
--- a/int/api/websites/webcreator.go
+++ b/int/api/websites/webcreator.go
@@ -1,7 +1,10 @@
 package websites
 
 import (
+	"archive/zip"
+	"bytes"
 	"encoding/base64"
+	"fmt"
 	"io"
 	"net/http"
 	"os"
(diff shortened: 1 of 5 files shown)


int/api/websites/webcreator.go
```diff
@@ -55,6 +82,18 @@ func prepareForWebsiteHandler(params operations.WebsiteCreatorPrepareParams, app
...
+	// Read all the files from zip archive
+	for _, zipFile := range zipReader.File {
+		if err != nil {
+			return nil
+		}
+		list_of_files = append(list_of_files, zipFile.Name)
+	}
+	// check if zip archive exist
+	if !contains(list_of_files, "index.html") {
+		return createInternalServerError(errorCodeWebCreatorHtmlNotInSource, err.Error())
```
Thykof, line 95:
```suggestion
		return createInternalServerError(errorCodeWebCreatorHTMLNotInSource, err.Error())
```

Thykof (changes requested):
There are some linter issues

B-Naoufal:
@Thykof just committed the suggestion modifications

Thykof (changes requested):
there is a changed file when I run `go generate` I think we should commit it

B-Naoufal:
> 

Done missed go generate for embedded spec

Thykof (changes requested):

Thykof (approved):

thomas-senechal (approved):

int/api/websites/webcreator.go
```diff
@@ -28,6 +30,25 @@ func CreatePrepareForWebsiteHandler(
 	}
 }
 
+func listArchiveContent(zipReader *zip.Reader) []string {
```
gregLibert, line 33: You function name is misleading, you should name it following what the code is doing.

My proposal:
```suggestion
func listFileName(zipReader *zip.Reader) []string {
```

int/api/websites/webcreator.go
```diff
@@ -28,6 +30,25 @@ func CreatePrepareForWebsiteHandler(
...
+
+func contains(elems []string, v string) bool {
+	for _, s := range elems {
+		if v == s {
+			return true
+		}
+	}
+
+	return false
+}
```
gregLibert, line 50: The best code is the one you're not writing : https://pkg.go.dev/golang.org/x/exp/slices#Index

int/api/websites/webcreator.go
```diff
@@ -61,8 +82,11 @@ func prepareForWebsiteHandler(params operations.WebsiteCreatorPrepareParams, app
 		return createInternalServerError(errorCodeWebCreatorArchiveSize, errorCodeWebCreatorArchiveSize)
 	}
 
-	if !checkContentType(archive, "application/zip") {
-		return createInternalServerError(errorCodeWebCreatorFileType, errorCodeWebCreatorFileType)
+	zipReader, _ := zip.NewReader(bytes.NewReader(archive), int64(len(archive)))
+	FilesOfArchive := listArchiveContent(zipReader)
+
+	if !contains(FilesOfArchive, "index.html") {
```
gregLibert, line 88: should be 
```suggestion
	if slice.Index(FilesOfArchive, "index.html") == -1 {
```

gregLibert (changes requested):
:no_good: 
You should use the native function instead of creating your own find in array.

The small suggestion for code clarity is more nitpicking (I would not have shared it without another comment).

gregLibert (approved):
👏 good job!

PR #283, opened 2022-11-21
remove uploaded date and import metadata to front

SebastienValla:
please check issue to see why we choosed to remove the date


int/api/websites/all.go
```diff
@@ -80,24 +74,25 @@ func Registry(client *node.Client, candidateDatastoreKeys [][]byte) ([]*models.R
...
-	for index, date := range dates {
+	for index := 0; index < len(metadatas); index++ {
 		registryResult[index] = &models.Registry{
-			Name:      strings.Split(recordKeys[index], recordKey)[1],
-			Address:   metadataKeys[index].Address,
-			CreatedAt: time.Unix(date.CreateDate/secondsToMilliCoeff, 0).Format(dateFormat),
-			UpdatedAt: time.Unix(date.UpdateDate/secondsToMilliCoeff, 0).Format(dateFormat),
+			Name:     strings.Split(recordKeys[index], recordKey)[1],
+			Address:  metadataKeys[index].Address,
+			Metadata: ByteArrToString(metadatas[index].CandidateValue),
```
gregLibert, line 83: If you want to convert it to string, why not using string(metadatas[index].CandidateValue[:])

int/api/websites/all.go
```diff
@@ -80,24 +74,25 @@ func Registry(client *node.Client, candidateDatastoreKeys [][]byte) ([]*models.R
...
 
-	dates := make([]dateOnChain, len(metadatas))
-
-	for index, metadata := range metadatas {
-		var date dateOnChain
-
-		_ = json.Unmarshal(metadata.CandidateValue, &date)
-		dates[index] = date
-	}
+	registryResult := make([]*models.Registry, len(metadatas))
```
gregLibert, line 77: Why using an array of pointers to models.Registry ?

int/api/websites/all.go
```diff
@@ -80,24 +74,25 @@ func Registry(client *node.Client, candidateDatastoreKeys [][]byte) ([]*models.R
...
+			Name:     strings.Split(recordKeys[index], recordKey)[1],
+			Address:  metadataKeys[index].Address,
+			Metadata: ByteArrToString(metadatas[index].CandidateValue),
 		}
 	}
 
 	return registryResult, nil
 }
+
+// transforms the []byte to string.
```
gregLibert, line 90: The coment is not matching the code. I don't understand why you add a coma between each char...

ASAPSegfault (approved):

gregLibert (changes requested):
:no_good:
I don't understand the business need but I challenge the technical choice:
- why using pointer instead of actual value to fill an array, as both array and underlying objects are created on purpose
- you are creating a function's logic that doesn't match its name and comment

int/api/html/front/registry.html
```diff
@@ -64,7 +64,6 @@ <h2 class="mb-4 mt-5">Browse decentralized websites</h2>
                     <table id="website-deployers-table" class="table table-striped">
                         <thead>
                             <tr>
-                                <th scope="col">Creation date</th>
```
Thykof, line 67: if the code is removed for work around, we should add a comment to explain it

int/api/websites/all.go
SebastienValla: I chosed to use a custom function and not string because string is the equivalent of fromCharCode and will return a string created from the specified sequence of UTF-16 code units. (expected output: "½+¾=") I wanted to have an output like this : 108,109,13,194,142,194,132,1,0,0,195 to be easier to handle on front end side and avoid to be reprocessed with charCodeAt()

SebastienValla: changed

int/api/websites/all.go
SebastienValla: changed

Thykof (approved):

gregLibert (approved):
:man_shrugging: LGTM

PR #286, opened 2022-11-22
Migrate API spec to OpenAPI 3

aoudiamoncef:
Hi,

OpenAPI 3 is massively adopted with a great set of tools around it. 

As you have a contract first approach, you could try to use [ApiCurio](https://studio.apicur.io/) to edit the API spec with a collaborative and user friendly GUI instead of editing a yaml file. 

IMHO,if you choose the solution above for you next changes, it's better to save the spec in JSON format.

I'm waiting for feedback

image

### DIFF
--- a/api/swagger/server/restapi/resource/swagger.yml
+++ b/api/swagger/server/restapi/resource/swagger.yml
@@ -1,728 +1,917 @@
-consumes:
-  - application/json
+openapi: 3.0.3
 info:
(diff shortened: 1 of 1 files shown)


gregLibert:
Thank you @aoudiamoncef for your suggestion. Unfortunetly the framework used to generate server side API code doesn't support openAPI 3.0.

PR #306, opened 2022-11-24
Load Wallet needs to be doable only once with same private key.

B-Naoufal:


pkg/wallet/wallet.go
```diff
@@ -192,11 +202,23 @@ func Imported(nickname string, privateKey string) (*Wallet, error) {
...
+	}
+
 	keypair := ed25519.NewKeyFromSeed(privKeyBytes)
 
 	pubKeyBytes := reflect.ValueOf(keypair.Public()).Bytes() // force conversion to byte array
 
 	addr := blake3.Sum256(pubKeyBytes)
+	Address := "A" + base58.CheckEncode(append(make([]byte, 1), addr[:]...))
+
+	if slices.Index(ListOfAddresses, Address) != -1 {
```
peterjah, line 219: I propose to use to remove GetListOfAddresses function and to use:

```slices.IndexFunc(wallets, func(wallet Wallet) bool { return wallet.Address == Address })```

it will iterate on the Wallet slice and test if any match with the provided Address

File: int/api/wallet/import.go [hunk 2]
```diff
@@ -61,11 +62,19 @@ func (c *wImport) Handle(params operations.MgmtWalletImportParams) middleware.Re
 
 	newWallet, err := wallet.Imported(walletName, privateKey)
 	if err != nil {
-		return operations.NewMgmtWalletCreateInternalServerError().WithPayload(
-			&models.Error{
-				Code:    errorCodeWalletCreateNew,
-				Message: err.Error(),
-			})
+		if strings.HasPrefix(err.Error(), "wallet already imported") {
```
peterjah, line 65: Comparing the error message is not a good practice. if we modify the error message your code will not work anymore.
Here you must create a custom error in the "imported" function to add an error code that you can compare.
you can see an example here under "Collecting Detailed Information in a Custom Error" section
https://www.digitalocean.com/community/tutorials/creating-custom-errors-in-go

peterjah (changes requested):

int/api/wallet/import.go
```diff
@@ -59,13 +59,34 @@ func (c *wImport) Handle(params operations.MgmtWalletImportParams) middleware.Re
...
+					Code:    errorCodeWalletLoadingWallets,
+					Message: RequestError.Err.Error(),
+				})
+		default:
+			return operations.NewMgmtWalletCreateInternalServerError().WithPayload(
+				&models.Error{
+					Code:    errorCodeWalletCreateNew,
+					Message: RequestError.Err.Error(),
+				})
+		}
```
Thykof, line 89: I have the feeling here that we duplicate a lot of code. I think that `wallet.Imported` should return the error code or nil.

```suggestion
		return operations.NewMgmtWalletCreateInternalServerError().WithPayload(
				&models.Error{
					Code:    returnedErrorCode,
					Message: RequestError.Err.Error(),
				})
```

int/api/wallet/import.go
```diff
@@ -59,13 +59,34 @@ func (c *wImport) Handle(params operations.MgmtWalletImportParams) middleware.Re
...
+				&models.Error{
+					Code:    errorCodeWalletLoadingWallets,
+					Message: RequestError.Err.Error(),
+				})
+		default:
+			return operations.NewMgmtWalletCreateInternalServerError().WithPayload(
+				&models.Error{
+					Code:    errorCodeWalletCreateNew,
+					Message: RequestError.Err.Error(),
+				})
```
peterjah, line 88: can you factorize this ?! something like NewWalletError( code, RequestError)

pkg/wallet/wallet.go
```diff
@@ -181,19 +220,49 @@ func New(nickname string) (*Wallet, error) {
...
+	WalletFromKeys, err := CreateWalletFromKeys(nickname, privKeyBytes, pubKeyBytes, addr)
+	if err != nil {
+		return nil, RequestError{
+			StatusCode: ImportedCreateWalletFromKeysError,
+			Err:        fmt.Errorf("create wallet from keys error  %w", err),
+		}
+	}
 
-	return CreateWalletFromKeys(nickname, privKeyBytes, pubKeyBytes, addr)
+	return WalletFromKeys, RequestError{StatusCode: ImportedNoError, Err: nil}
```
peterjah, line 265: why you do not just return WalletFromKeys, nil

peterjah (changes requested):

int/api/wallet/import.go
B-Naoufal: cannot use nil as RequestError value, since type RequestError struct {
	StatusCode int
	Err        error
}

int/api/wallet/import.go
Thykof: okay, is there a way that `wallet.Imported` returns `operations.NewMgmtWalletCreateInternalServerError`?

B-Naoufal: @Thykof it lead us to the first issue : cyclic import

B-Naoufal: so we cannot import just te errors which will be used by operations.NewMgmtWalletCreateInternalServerError
image

B-Naoufal: @gregLibert do you have a better suggestion to avoid cycle import errors ? i used a custom errors inside pkg/wallet/wallet.go and a switch to match my return errors in int/api/wallet/errors.go since i cannot import int/api/wallet/errors.go inside pkg/wallet. but it seems its not very optimal solution

Thykof: is it possible to move errors.go to solve cyclic import?

Thykof:
it's better with commit 15f2edb80948f192f746a0326fa5662aa0b81f94 but I still have the feeling that we have a useless passthrough

peterjah (approved):

int/api/wallet/import.go
```diff
@@ -44,29 +41,30 @@ func (c *wImport) Handle(params operations.MgmtWalletImportParams) middleware.Re
...
 	}
 
-	newWallet, err := wallet.Imported(walletName, privateKey)
-	if err != nil {
-		return operations.NewMgmtWalletCreateInternalServerError().WithPayload(
-			&models.Error{
-				Code:    errorCodeWalletCreateNew,
-				Message: err.Error(),
-			})
+	newWallet, Err := wallet.Imported(walletName, privateKey)
```
Thykof, line 51: variable `Err` can be lowercase.

Thykof (changes requested):
Great work! one last variable naming code smel

Thykof (approved):
🤷 LGTM

int/api/wallet/import.go
```diff
@@ -44,29 +41,30 @@ func (c *wImport) Handle(params operations.MgmtWalletImportParams) middleware.Re
...
 
 	newWallet, err := wallet.Imported(walletName, privateKey)
 	if err != nil {
-		return operations.NewMgmtWalletCreateInternalServerError().WithPayload(
-			&models.Error{
-				Code:    errorCodeWalletCreateNew,
-				Message: err.Error(),
-			})
+		if errors.Is(err, wallet.ErrWalletAlreadyImported) {
+			return NewWalletError(errorCodeWalletAlreadyImported, wallet.ErrWalletAlreadyImported.Error())
```
gregLibert, line 54:
```suggestion
			return NewWalletError(errorCodeWalletAlreadyImported, err.Error())
```

int/api/wallet/import.go
```diff
@@ -44,29 +41,30 @@ func (c *wImport) Handle(params operations.MgmtWalletImportParams) middleware.Re
...
 }
+
+//nolint:nolintlint,ireturn
+func NewWalletError(errorCode string, errorMessage string) middleware.Responder {
+	return operations.NewMgmtWalletCreateInternalServerError().WithPayload(
+		&models.Error{
+			Code:    errorCode,
+			Message: errorMessage,
+		})
+}
```
gregLibert, line 70:
```suggestion
func NewWalletError(code string, message string) middleware.Responder {
	return operations.NewMgmtWalletCreateInternalServerError().WithPayload(
		&models.Error{
			Code:    code,
			Message: message,
		})
}
```

pkg/wallet/wallet.go
```diff
@@ -187,11 +191,23 @@ func Imported(nickname string, privateKey string) (*Wallet, error) {
...
+	wallets, err := LoadAll()
+	if err != nil {
+		return nil, fmt.Errorf("error loadin wallets %w", err)
+	}
 
+	keypair := ed25519.NewKeyFromSeed(privKeyBytes)
 	pubKeyBytes := reflect.ValueOf(keypair.Public()).Bytes() // force conversion to byte array
 
 	addr := blake3.Sum256(pubKeyBytes)
+	Address := "A" + base58.CheckEncode(append(make([]byte, 1), addr[:]...))
```
gregLibert, line 203: Can't you use the versioned version of check encode ?

gregLibert (approved):
:mag:  readability of the code

pkg/wallet/wallet.go
```diff
@@ -187,11 +191,24 @@ func Imported(nickname string, privateKey string) (*Wallet, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("error loadin wallets %w", err)
+	}
 
+	keypair := ed25519.NewKeyFromSeed(privKeyBytes)
 	pubKeyBytes := reflect.ValueOf(keypair.Public()).Bytes() // force conversion to byte array
 
 	addr := blake3.Sum256(pubKeyBytes)
+	//nolint:gomnd
+	Address := "A" + base58.VersionedCheckEncode(addr[:] /* current version is 0*/, 0x00)
```
gregLibert, line 204: The linter is yelling, I propose to:
- create an explaining variable
- remove the now useless comment

Why are you using an upper case A for the address variable?
```suggestion
	version := 0x00
	address := "A" + base58.VersionedCheckEncode(addr[:], 0x00)
```

gregLibert (changes requested):
Sorry, I didn't think about the linter. And I also saw another mistake :-)

gregLibert (approved):
:+1:

PR #330, opened 2022-12-08
Read file as binary instead of b64 string

ASAPSegfault:
The chunk size was computed with b64 encoded data. Now it's computed with binary data to match our new model.


Thykof (approved):
LGTM 🤷‍♂️

0xMazout (approved):

SebastienValla (approved):

PR #332, opened 2022-12-08
new dns contract

SebastienValla:


0xMazout (approved):
LETS GOO MERGE

Thykof (approved):
We need to find a way to avoid to commit the wasm file here, like compile it in massa-as-contract, create a release with wasm file as artifact and here fetch it.

B-Naoufal (approved):
For me, it looks sufficient.

gregLibert (approved):
:camping: :clap:

PR #354, opened 2022-12-10
refactor install scripts

peterjah:


Thykof:
@thomas-senechal can you confirm that .sh script are not used anymore? if it's true then let's remove them.

SlnPons:
What is the status of this issue please? 
I see that it is blocked by the #503. But I can't see any PR linked to it... so when #503 is done, do we need to do this issue or is it already done and we need to review it?

Thykof:
this PR is modifying deprecated installation files for thyra. @thomas-senechal wants to wait for https://github.com/massalabs/thyra/pull/518 to be merge before removing those useless files. Overall I think we can close this PR. @peterjah what do you think?

PR #358, opened 2022-12-12
345 install script for windows in python

ASAPSegfault:


scripts/thyra-installer.py
```diff
@@ -0,0 +1,132 @@
+import urllib.request
+import subprocess
+import shutil
+import os
+import zipfile
+import ctypes
+from urllib.error import URLError
+from tarfile import ReadError
+
+# This file is to be bundled with pyinstaller in order to produce a .exe that can run on Windows without Python installed.
```
gregLibert, line 10:
```suggestion
# TODO: This file is to be bundled with pyinstaller in order to produce a .exe that can run on Windows without Python installed.
```

scripts/thyra-installer.py
```diff
@@ -0,0 +1,132 @@
...
+
+def executeOSCommandOrFile(command, decodeBinary, errorChecking=True):
+    process = subprocess.Popen(command,
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE,
+                            universal_newlines=decodeBinary)
+    stdout, stderr = process.communicate()
+
+    if stderr != None and stderr != "" and errorChecking == True:
+        print("Error encountered while executing : " + str(command) + " :\n", stderr)
```
gregLibert, line 55:
```suggestion
        print("Error encountered while executing " + str(command) + " :\n", stderr)
```

scripts/thyra-installer.py
```diff
@@ -0,0 +1,132 @@
...
+        os._exit(-1)
```
gregLibert, line 56: Why not using printErrorAndExit here like before?

scripts/thyra-installer.py
```diff
@@ -0,0 +1,132 @@
...
+        printErrorAndExit("Couldn't detect admin rights. Please execute this script as an administator.")
+    downloadFile(THYRA_URL, THYRA_FILENAME)
+    if os.path.exists(DEFAULT_ACRYLIC_PATH):
+        print("Acrylic DNS Proxy is already installed")
+    else:
+        downloadFile(ACRYLIC_DNS_PROXY_URL, ACRYLIC_DNS_PROXY_FILENAME)
+        unzipAcrylic()
+        executeOSCommandOrFile(DEFAULT_ACRYLIC_PATH + "\InstallAcrylicService.bat", True)
+        setupDNS()
+    configureAcrylic()
```
gregLibert, line 126: What if Acrylic is already configured ? If it's safe, we can live with that.

scripts/thyra-installer.py
```diff
@@ -0,0 +1,132 @@
...
+    setupMkCerts()
```
gregLibert, line 127: What if we aleady have a CA certificate ? If it's safe, we can live with that !

gregLibert (changes requested):
:clap: A huge improvement!
:mag: suggestion to improve readibility
:thinking: and open questions :-)

scripts/thyra-installer.py
ASAPSegfault: This is not something that is left to do, it's already done, the .exe file contained in the PR is the bundled file.
I put this comment just to emphasize the fact that this python file is not meant to be used on its own.
Maybe I should explicitly state in the comment that in order to make a new release of our automatic installation process for Windows, this operation needs to be repeated to produce an updated .exe file.

scripts/thyra-installer.py
ASAPSegfault: Here's what is happening in the configureAcrylic() function :
- First I check if "127.0.0.1 *.massa" is contained in the AcrylicHosts file.
If it does, I do nothing (I just restart the Acrylic service, which is useless and I'll update this)
If it doesn't, I append it to the file and restart the Acrylic service.

So, if Acrylic is already "configured" => meaning here that "127.0.0.1 *.massa" is already present in the file, nothing happens anyway (besides the service restart).

scripts/thyra-installer.py
ASAPSegfault: No reason I forgot, will update

scripts/thyra-installer.py
peterjah: maybe you can check the return code of mkcert ?
 https://stackoverflow.com/questions/334879/how-do-i-get-the-application-exit-code-from-a-windows-command-line

peterjah:
Why are you commiting a .exe binary here ??? thats not acceptable to me.
This binary must be produced by the CI and published as artifact + in releases artifacts.

peterjah:
for the CI you may use this github action: 
https://github.com/marketplace/actions/pyinstaller-windows

scripts/thyra-installer.py
gregLibert: Understood. But I think we should use the CD script to create a binary file on the fly.

gregLibert (changes requested):

scripts/thyra-installer.py
ASAPSegfault: I ended up checking the return code, which is 0 as expected.

.github/workflows/build.yml
```diff
@@ -66,11 +66,22 @@ jobs:
         run: echo -e "${GITHUB_REF_NAME}" > cmd/thyra-server/version.txt
       - name: Build binary for ${{ matrix.target }} on ${{ matrix.arch}}
         run: GOOS=${{ matrix.target }} GOARCH=${{ matrix.arch }} CGO_ENABLED=1 CC=${{ matrix.cc }} CXX=${{ matrix.cxx }} go build -o thyra-server_${{ matrix.target }}_${{ matrix.arch }} ./cmd/thyra-server/
+      - name: Build installer binary
+        if: ${{ matrix.os == 'ubuntu-latest' }}
```
peterjah, line 70: suggestion:
`if: ${{ matrix.target == 'windows' }}`
could help to understand easier !

.github/workflows/build.yml
```diff
@@ -66,11 +66,22 @@ jobs:
...
+        uses: JackMcKew/pyinstaller-action-windows@main
+        with:
+          path: scripts
       - name: Upload artifacts
         uses: actions/upload-artifact@v3
         with:
           name: ${{ matrix.target }}-${{ matrix.arch }}-binary
           path: thyra-server_${{ matrix.target }}_${{ matrix.arch }}
+      - name: Upload installer artifacts
+        if: ${{ matrix.os == 'ubuntu-latest' }}
```
peterjah, line 80: same!

gregLibert (approved):
:tada: so cool!

0xMazout (approved):

.github/workflows/build.yml
```diff
@@ -66,11 +66,36 @@ jobs:
...
+        if: ${{ matrix.target == 'windows' }}
+        uses: JackMcKew/pyinstaller-action-windows@main
+        with:
+          path: scripts
+      - name: Package installer folder into Windows executable file
+        if: ${{ matrix.target == 'windows' }}
+        run: |
+          curl -L https://www.7-zip.org/a/lzma2201.7z --output lzma2201.7z
+          p7zip -d lzma2201.7z
+          echo -e ';!@Install@!UTF-8!\n
```
peterjah, line 79: I think it will be easier if you create this config.txt file in the repo and commit it

PR #359, opened 2022-12-16
307 certificate process

0xMazout:


0xMazout:
image

0xMazout:
@gregLibert

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
+    // Utilisez IntelliSense pour en savoir plus sur les attributs possibles.
+    // Pointez pour afficher la description des attributs existants.
+    // Pour plus d'informations, visitez : https://go.microsoft.com/fwlink/?linkid=830387
```
thomas-senechal, line 4: I don't think we need those comments.

api/swagger/server/restapi/certHandler.go
```diff
@@ -0,0 +1,222 @@
...
+		dir = os.Getenv("LocalAppData")
+	case os.Getenv("XDG_DATA_HOME") != "":
+		dir = os.Getenv("XDG_DATA_HOME")
+	case runtime.GOOS == "darwin":
+		dir = os.Getenv("HOME")
+		if dir == "" {
+			return ""
+		}
+		dir = filepath.Join(dir, "Library", "Application Support")
+	default: // Unix
```
thomas-senechal, line 96: I don't think it's a good idead to use `default` for "`Unix`" since it might be something else than a `Unix` OS.

api/swagger/server/restapi/certHandler.go
```diff
@@ -0,0 +1,222 @@
...
+	switch {
+	case runtime.GOOS == "windows":
+		dir = os.Getenv("LocalAppData")
+	case os.Getenv("XDG_DATA_HOME") != "":
+		dir = os.Getenv("XDG_DATA_HOME")
+	case runtime.GOOS == "darwin":
+		dir = os.Getenv("HOME")
+		if dir == "" {
+			return ""
+		}
```
thomas-senechal, line 94: I think we should return an `error` instead of returning an empty string. It might be confusing to receive an empty string. Especially since this behavior isn't documented.

api/swagger/server/restapi/certHandler.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// Get the CA root key or create a new one
+func (m *mkcert) loadCA() {
+	if !pathExists(filepath.Join(m.CAROOT, rootName)) {
+		m.newCA()
+	}
+
+	certPEMBlock, err := os.ReadFile(filepath.Join(m.CAROOT, rootName))
+	// fmt.Println(err, "failed to read the CA certificate")
```
thomas-senechal, line 113: I don't think this comment is useful. 
```suggestion
```

api/swagger/server/restapi/configure_thyra_server.go
```diff
@@ -63,8 +62,27 @@ func configureAPI(api *operations.ThyraServerAPI) http.Handler {
...
+func myCert(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	certBytes, priv, err := (&mkcert{}).Run(hello.ServerName)
+	var cert tls.Certificate
+	if err != nil {
+		return nil, err
+	}
+	cert.Certificate = append(cert.Certificate, certBytes)
+	cert.PrivateKey = priv
+
+	fmt.Printf("%v", hello)
```
thomas-senechal, line 74: Is this print useful ? 
If yes, I think you should use: 
```suggestion
	log.Printf("%v", hello)
```

api/swagger/server/restapi/configure_thyra_server.go
```diff
@@ -84,9 +102,10 @@ func configureTLS(tlsConfig *tls.Config) {
 		var err error
 
 		tlsConfig.Certificates = make([]tls.Certificate, 1)
-		tlsConfig.Certificates[0], err = tls.X509KeyPair(unsecureCertificate, unsecureKey)
-
+		certiff, err := tls.X509KeyPair(unsecureCertificate, unsecureKey)
+		tlsConfig.Certificates = append(tlsConfig.Certificates, certiff)
 		if err != nil {
+			fmt.Println("error: unable to load certificate and key files.")
```
thomas-senechal, line 108: I think you should use `log` instead of
```suggestion
			log.Println("error: unable to load certificate and key files.")
```

api/swagger/server/restapi/certHandler.go
```diff
@@ -0,0 +1,222 @@
...
+	return serialNumber
+}
+
+// return true if path exist
+func pathExists(path string) bool {
+	_, err := os.Stat(path)
+	return err == nil
+}
+
+func fatalIfErr(err error, msg string) {
```
thomas-senechal, line 218: I'm not sure about this function. I don't know if we gain or lose in code readability.
I'd like the opinion of someone else about it. 

I think we kinda hide what we are doing and that, even if the function name is really clear and that it simplifies the code a lot, I'm not sure it's a good thing to hide this check.

If we decide to keep it, I think we should put it somewhere else to use it everywhere in Thyra to have a consistant code.

thomas-senechal (changes requested):

api/swagger/server/restapi/certHandler.go
thomas-senechal: If we add a `case` for `linux` (I don't think `Unix` exists since MacOS is a member of the `Unix` family, as Linux, Android, BSD... [source](https://fr.wikipedia.org/wiki/Unix)), I think the `default` should return an error as explained in [my other comment](https://github.com/massalabs/thyra/pull/359#discussion_r1052353710).

0xMazout: I think this is useless because , there won't be other choice. This is using fileDirectory from specific OS and there is no others OS  which can be used with this process. There is no Default because we set all the OS available. 
For me it's useless

api/swagger/server/restapi/certHandler.go
0xMazout: Don't agree we return a string if the string is empty the home is not set i added a commentary. This is more easy . And the error is handle on the caller.

0xMazout:
MKCERT link : 
https://github.com/FiloSottile/mkcert/blob/master/main.go

Contains : 
GetCaRoot
generateKey
randomSerialNumber
pathExists
newCA
loadCA

thomas-senechal (approved):

peterjah (changes requested):
Great achievement !
As i understand we do not need to generate a certificate for my.massa with mkcert at install. thus, can you clean install scripts ?!

api/swagger/server/restapi/configure_thyra_server.go
```diff
@@ -63,8 +63,26 @@ func configureAPI(api *operations.ThyraServerAPI) http.Handler {
 //go:embed resource
 var content embed.FS
 
+func myCert(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	certBytes, priv, err := (&mkcert{}).Run(hello.ServerName)
+	var cert tls.Certificate
+	if err != nil {
+		return nil, err
+	}
```
Thykof, line 71: variable declaration should be as close as possible to where it is used 

```suggestion
	if err != nil {
		return nil, err
	}
	var cert tls.Certificate
```

api/swagger/server/restapi/configure_thyra_server.go
```diff
@@ -63,8 +63,26 @@ func configureAPI(api *operations.ThyraServerAPI) http.Handler {
...
+	cert.Certificate = append(cert.Certificate, certBytes)
+	cert.PrivateKey = priv
+
+	return &cert, nil
+}
+
 // The TLS configuration before HTTPS server starts.
 func configureTLS(tlsConfig *tls.Config) {
+	var err error
```
Thykof, line 80: this line may be useless because line 88 declare `err` with `:=`

Thykof (changes requested):

api/swagger/server/restapi/certHandler.go
```diff
@@ -0,0 +1,226 @@
+package restapi
+
```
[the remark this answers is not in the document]
0xMazout: That was my first purpose , it would be cleaner but in mkcert none of the functions are public :/
And we were thinking it's okay to take this because mkcert is not gonna to be a product super updated in time

api/swagger/server/restapi/configure_thyra_server.go
```diff
@@ -84,9 +102,10 @@ func configureTLS(tlsConfig *tls.Config) {
 		var err error
 
 		tlsConfig.Certificates = make([]tls.Certificate, 1)
-		tlsConfig.Certificates[0], err = tls.X509KeyPair(unsecureCertificate, unsecureKey)
-
+		certiff, err := tls.X509KeyPair(unsecureCertificate, unsecureKey)
```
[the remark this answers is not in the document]
0xMazout: if we can't  load a certificate with generateCertificate , this will be our exit door ?

Or Maybe i'm wrong and this won't happend any time

0xMazout:
> Great achievement ! As i understand we do not need to generate a certificate for my.massa with mkcert at install. thus, can you clean install scripts ?!

Yes i think we can clean this

Thykof (approved):
LGTM

api/swagger/server/restapi/certHandler.go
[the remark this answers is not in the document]
0xMazout: Maybe i was not clear.

Mkcert is a tool with unique goal. And you can't pick what you want on it. You call Run and thats it. And i didn't need all of the tool and my using was not aligned with the goal of mkcert

api/swagger/server/restapi/certHandler.go
```diff
@@ -0,0 +1,226 @@
+package restapi
```
gregLibert, line 1: This code should not be here. You should put it in:
- /int (for /internal) directory if it's not ready to be an external package yet
- /pkg if you think that the code can be reused as is by another project.

api/swagger/server/restapi/certHandler.go
```diff
@@ -0,0 +1,226 @@
...
+	"io/ioutil"
+	"log"
+	"math/big"
+	"os"
+	"path/filepath"
+	"runtime"
+	"time"
+)
+
+var userAndHostname string
```
gregLibert, line 24: Why do you give this variable such a big scope ?

api/swagger/server/restapi/configure_thyra_server.go
```diff
@@ -63,8 +63,25 @@ func configureAPI(api *operations.ThyraServerAPI) http.Handler {
...
+	if err != nil {
+		return nil, err
+	}
+	var cert tls.Certificate
+	cert.Certificate = append(cert.Certificate, certBytes)
+	cert.PrivateKey = priv
+
+	return &cert, nil
+}
+
```
gregLibert, line 77: this should moved somewhere else.
The configure_thyra_server.go file is generated by go swagger. This means that we will potentially need to regenerate it at some point. Identifiying what is automaticaly generated and what was added manualy will be painful. hence my remarks...

api/swagger/server/restapi/configure_thyra_server.go
```diff
@@ -63,8 +63,25 @@ func configureAPI(api *operations.ThyraServerAPI) http.Handler {
...
 // The TLS configuration before HTTPS server starts.
 func configureTLS(tlsConfig *tls.Config) {
+
+	var helloInfo tls.ClientHelloInfo
+	helloInfo.ServerName = tlsConfig.ServerName
+
+	tlsConfig.GetCertificate = generateCertificate
 	basePath := "resource/certificate/"
```
gregLibert, line 85: This is not to use anymore and then should be removed.

api/swagger/server/restapi/configure_thyra_server.go
```diff
@@ -63,8 +63,25 @@ func configureAPI(api *operations.ThyraServerAPI) http.Handler {
...
+	cert.PrivateKey = priv
+
+	return &cert, nil
+}
+
 // The TLS configuration before HTTPS server starts.
 func configureTLS(tlsConfig *tls.Config) {
+
+	var helloInfo tls.ClientHelloInfo
+	helloInfo.ServerName = tlsConfig.ServerName
```
gregLibert, line 82: Why do you need that ?

api/swagger/server/restapi/configure_thyra_server.go
```diff
@@ -63,8 +63,25 @@ func configureAPI(api *operations.ThyraServerAPI) http.Handler {
 //go:embed resource
 var content embed.FS
 
+func generateCertificate(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	certBytes, priv, err := (&mkcert{}).Run(hello.ServerName)
```
gregLibert, line 67: private key should be static and shared for all website.

api/swagger/server/restapi/certHandler.go
```diff
@@ -0,0 +1,226 @@
...
+
+const rootName = "rootCA.pem"
+const rootKeyName = "rootCA-key.pem"
+
+type mkcert struct {
```
gregLibert, line 29: you don't need that. here you copy/past mkcert code without removing what you really need.
Do you need help to clean all that ?

gregLibert (changes requested):
:no_good: To much uneeded code here. I can help if necessary.

api/swagger/server/restapi/certHandler.go
0xMazout: I need this , this is the name of my CA certificate and my key CA 

We can rename it massaCa or something like that but i need this to load the CA

pkg/certificate/x509.go
```diff
@@ -0,0 +1,147 @@
...
+func mkcertCARootPath() (string, error) {
+	if env := os.Getenv("CAROOT"); env != "" {
+		return env, nil
+	}
+
+	var dir string
+
+	switch {
+	case runtime.GOOS == "windows":
+		dir = os.Getenv("LocalAppData")
```
ASAPSegfault, line 77: How would this behave on a computer setup in Chinese/Indian/Arabic ? Does this environment variable exists on all Windows distributions ?

pkg/certificate/x509.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, nil, fmt.Errorf("while loading CA certificate: %w", err)
+	}
+
+	caPrivateKey, err := LoadPrivateKey(caPath + "/" + rootKeyName)
+	if err != nil {
+		return nil, nil, fmt.Errorf("while loading CA key: %w", err)
+	}
+
+	//nolint:exhaustruct
+	tpl := &x509.Certificate{
```
ASAPSegfault, line 45: I think this variable should be renamed because tpl is not straightforward to understand

pkg/certificate/x509.go
gregLibert: I think so but I'm not sure (windows :shrug:).  In anycase, this is exactly what mkcert is doing and as mkcert is the current source of the CA...

pkg/certificate/x509.go
gregLibert: changed to template

pkg/certificate/x509.go
ASAPSegfault: Fair enough

ASAPSegfault (approved):

gregLibert (approved):

PR #360, opened 2022-12-19
280 website registry link does not take thyra running port into account

thomas-senechal:


peterjah (approved):

peterjah:
we will have to change again theses links for the support of https ! to be continued !

thomas-senechal:
> we will have to change again theses links for the support of https ! to be continued !

I'm not sure about that. We can remove the "http:" part and the web browser will figure out by itself which protocol it should use. By default, it will try to use `https`. 

I just tried it and it's working fine. I will push that fix.

int/api/html/front/registry.js
```diff
@@ -24,7 +24,7 @@ function tableInsert(website) {
         .getElementById("website-deployers-table")
         .getElementsByTagName("tbody")[0];
     const row = tBody.insertRow(-1);
-    const url = "http://" + website[NAME] + ".massa/";
+    const url = `//${website[NAME]}.massa:${window.location.port}/`;
```
gregLibert, line 27:
```suggestion
    const port = window.location.port == 80 ? "" : ":" + window.location.port;
    const url = `${windows.location.protocol}//${website[NAME]}.massa${port}/`;
```

int/api/html/front/website.js
```diff
@@ -162,7 +162,7 @@ async function tableInsert(resp, count) {
         .getElementById("website-deployers-table")
         .getElementsByTagName("tbody")[0];
     const row = tBody.insertRow(-1);
-    const url = "http://" + resp.name + ".massa/";
+    const url = `//${resp.name}.massa:${window.location.port}/`;
```
gregLibert, line 165: idem

gregLibert (changes requested):
We will have https really soon :-)

int/api/html/front/registry.js
thomas-senechal: I'm not sure think we need to add the `protocol` at the beginning since

gregLibert (approved):
:+1:

PR #361, opened 2022-12-20
Updates the bug template

SlnPons:


.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -7,30 +7,18 @@ assignees: ''
...
 
 **Desktop (please complete the following information):**
  - OS: [e.g. MacOS, Windows...]
- - Browser [e.g. chrome, safari]
- - Version [e.g. 22]
-
-**Additional context**
-Add any other context about the problem here.
\ No newline at end of file
+ - Any other useful info
```
thomas-senechal: Since you merged the `Additional context` section into the `Desktop` section, I think we should rename it to `Additional context` since it includes the OS and any other information de user has to give us.

thomas-senechal (changes requested):
I'm not sure we are going in the right way with this PR. 
We are removing some information that are supposed to help the user to fill the issue description. 

Maybe instead of removing those information, we should use the `.yml` issue template format ([doc with example](https://docs.github.com/fr/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms)) to allow us to have clean form with all the informations a user might need to write an issue that will contains the information we need ? 
Since this template will be mainly used by end-users, I think it's worth to take the time to use that format for those kind of issues.

.github/ISSUE_TEMPLATE/bug.md
SlnPons: I'm generally not a huge fan of the "fourre tout" categories. But you had a point that it's not relevant into the `Desktop` section. So I modified it as you suggested

SlnPons:
> 

As discussed together, I won't address your comment. First because the previous bug template wasn't used <=> it was not right.
And secondly because we're still not at the stage where external users create bug so let's keep it as simple as possible so we have room for easy and quick improvement. The more we use it the better this template will be.

Thykof (approved):

gregLibert (approved):

PR #369, opened 2022-12-28
287 Hardcode the DNS addresses in the code and create a flag for handling multiple DNS

thomas-senechal:
Note that since we only have a DNS deployed on TestNet 17, we can't add default DNS for the others Networks.

### DIFF
--- a/int/api/server.go
+++ b/int/api/server.go
@@ -15,6 +15,7 @@ import (
 	"github.com/massalabs/thyra/int/api/wallet"
 	"github.com/massalabs/thyra/int/api/websites"
 	"github.com/massalabs/thyra/pkg/node"
+	"github.com/massalabs/thyra/pkg/onchain/dns"
 	pluginmanager "github.com/massalabs/thyra/pkg/plugins"
 )
 
@@ -24,6 +25,7 @@ type StartServerFlags struct {
 	TLSCertificate    string
 	TLSCertificateKey string
 	MassaNodeServer   string
+	DNSAddress        string
 	Version           bool
 }
 
@@ -40,12 +42,19 @@ func setAPIFlags(server *restapi.Server, startFlags StartServerFlags) {
 	}
 
 	parseNetworkFlag(&startFlags.MassaNodeServer)
+
+	if startFlags.DNSAddress != "" {
+		os.Setenv(dns.DNSAddressEnvKey, startFlags.DNSAddress)
+	}
 }
 
 func parseNetworkFlag(massaNodeServerPtr *string) {
+	var dnsAddressPtr string
+
 	switch *massaNodeServerPtr {
 	case "TESTNET":
 		*massaNodeServerPtr = "https://test.massa.net/api/v2"
+		dnsAddressPtr = "A15e47ChESAK1SdmGe3b92bybnBvMX2eFaxg23wn3rSdRzFHHGB"
 	case "LABNET":
 		*massaNodeServerPtr = "https://labnet.massa.net/api/v2"
 	case "INNONET":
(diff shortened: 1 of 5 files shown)


SlnPons:
@thomas-senechal I don't see any comment on the PR, could we add the label "action:merge"?

thomas-senechal:
> @thomas-senechal I don't see any comment on the PR, could we add the label "action:merge"?

Oops.
In fact, I just forgot to put that PR in review. 
We now have to wait for 2 reviews before adding the "action:merge" label.

int/api/server.go
```diff
@@ -54,6 +63,7 @@ func parseNetworkFlag(massaNodeServerPtr *string) {
 		*massaNodeServerPtr = "http://127.0.0.1:33035"
 	}
 
+	os.Setenv(dns.DNSAddressEnvKey, dnsAddressPtr)
```
gregLibert, line 66: You should avoid to use the package name in the constant. It's redundant.

int/api/websites/all.go
```diff
@@ -26,7 +26,7 @@ const (
 func RegistryHandler(params operations.AllDomainsGetterParams) middleware.Responder {
 	client := node.NewDefaultClient()
 
-	addressesResult, err := ledger.Addresses(client, []string{dns.DNSRawAddress})
+	addressesResult, err := ledger.Addresses(client, []string{dns.GetDNSAddress()})
```
gregLibert, line 29: Same here: Get and DNS are useless.

pkg/onchain/dns/dns.go
```diff
@@ -12,7 +13,11 @@ import (
 	"github.com/massalabs/thyra/pkg/wallet"
 )
 
-const DNSRawAddress = "A15e47ChESAK1SdmGe3b92bybnBvMX2eFaxg23wn3rSdRzFHHGB"
+const DNSAddressEnvKey = "THYRA_DNS_ADDRESS"
+
+func GetDNSAddress() string {
+	return os.Getenv(DNSAddressEnvKey)
+}
```
gregLibert, line 20:
```suggestion
const EnvKey = "THYRA_DNS_ADDRESS"

func Address() string {
	return os.Getenv(EnvKey)
}
```

gregLibert (changes requested):
:no_good: The readability of the code needs to be improved

int/api/server.go
```diff
@@ -40,20 +42,29 @@ func setAPIFlags(server *restapi.Server, startFlags StartServerFlags) {
...
 
 	parseNetworkFlag(&startFlags.MassaNodeServer)
+
+	if startFlags.DNSAddress != "" {
+		os.Setenv(dns.EnvKey, startFlags.DNSAddress)
+	}
 }
 
 func parseNetworkFlag(massaNodeServerPtr *string) {
+	var dnsAddressPtr string
```
gregLibert, line 52: What does Ptr mean ?

gregLibert (approved):
:thinking:

int/api/server.go
thomas-senechal: Fixed!

B-Naoufal (approved):

int/api/server.go
```diff
@@ -40,20 +42,29 @@ func setAPIFlags(server *restapi.Server, startFlags StartServerFlags) {
...
+	}
 }
 
 func parseNetworkFlag(massaNodeServerPtr *string) {
+	var dnsAddress string
+
 	switch *massaNodeServerPtr {
 	case "TESTNET":
 		*massaNodeServerPtr = "https://test.massa.net/api/v2"
+		dnsAddress = "A15e47ChESAK1SdmGe3b92bybnBvMX2eFaxg23wn3rSdRzFHHGB"
```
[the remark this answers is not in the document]
thomas-senechal: Discussed in PM

PR #372, opened 2022-12-29
344 possibility to delete websites on the registry page

SebastienValla:
As testnet is down it is to be tested on labnet  - > go run .\cmd\thyra-server\main.go -node-server LABNET


0xMazout (approved):
**Nice Work !**

thomas-senechal (approved):

PR #391, opened 2023-01-05
change serial number in certif

SebastienValla:


pkg/certificate/certificate.go
```diff
@@ -57,12 +58,12 @@ func GenerateTLS(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
 		return nil, wrapAndPrint("while getting mkcert CA root path", err)
 	}
 
-	caCertificate, err := LoadCertificate(caPath + "/rootCA.pem")
+	caCertificate, err := LoadCertificate(filepath.Join(caPath, "rootCA.pem"))
```
gregLibert, line 61: Good catch !

pkg/certificate/certificate.go
```diff
@@ -90,9 +91,12 @@ func GenerateSignedCertificate(
 		return nil, nil, errors.New("while generating certificate: server name is empty")
 	}
 
+	SHA256serverNameEncrypted := sha256.New().Sum([]byte(serverName))
```
gregLibert, line 94: The name of this variable only describes its technical content. It doesn't provide any functional or business information.

For example, if you name it websiteNameHash, it will make the reader understand that it is not the sha256 hash function that is important, but the fact of hashing the name of the website, to make it unique. This will also make this variable resistant to a potential change of the underlying hash function.

That being said, my suggestion is this:
```suggestion
	uniqueSiteNameID := sha256.New().Sum([]byte(serverName))
```

pkg/certificate/certificate.go
```diff
@@ -90,9 +91,12 @@ func GenerateSignedCertificate(
...
+
 	//nolint:exhaustruct
 	template := &x509.Certificate{
-		SerialNumber: &big.Int{}, // mandatory, but useless as it should be a unique id given to the certificate by the CA
+		//nolint:lll
+		SerialNumber: new(big.Int).SetBytes(SHA256serverNameEncrypted), // in order to have an unique ID for each website Name, we hash the website name
```
gregLibert, line 99: After the analysis, I wondering if this is really mandatory. Can you try, now that you have manually added the CA root certificate to you firefox, to use the previous code to see if it works or not ?

gregLibert (changes requested):

pkg/certificate/certificate.go
SebastienValla: not working whitout the modifs : Code d’erreur : SEC_ERROR_REUSED_ISSUER_AND_SERIAL

gregLibert: ok so we need to merge this PR. Thx !

gregLibert (approved):
:+1:

0xMazout (approved):
Nice !

PR #404, opened 2023-01-11
1 click install macos

thomas-senechal:
Latest binaries available here: https://github.com/massalabs/thyra/actions/runs/3939656469

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,18 +77,15 @@ jobs:
           files: thyra-server_${{ matrix.target }}_${{ matrix.arch }}
 
   build-installer:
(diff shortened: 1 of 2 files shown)


thomas-senechal:
If someone takes this issues while I'm off:
For now, the script should work properly for MacOS. But my changes might have break the Windows side. This **must** be tested on Windows before merge.
I didn't tried to make a MacOS installer with the updated script. I only tested it using `python3 script/thyra-installer.py`. 
You need also need to update the `build` workflow to build a MacOS installer as we did for Windows.

Thykof (commented):
scripts/thyra-installer.py is getting long, we might think about implementing strategy design pattern

PR #418, opened 2023-01-20
411 add wallet balance to thyra node object

SebastienValla:


Thykof (approved):
LGTM

pkg/node/getBalanceOf.go
```diff
@@ -0,0 +1,31 @@
...
+	"fmt"
+	"strconv"
+)
+
+type GetBalanceOfResponse struct {
+	CandidateBalance float64
+	FinalBalance     float64
+}
+
+// GetBalanceOf returns as a float64 the candidate balance and final balance of an address
```
Thykof, line 13:
```suggestion
// GetBalanceOf returns as a float64 the candidate balance and final balance of an address.
```

thomas-senechal (approved):

pkg/node/getBalanceOf.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"fmt"
+	"strconv"
+)
+
+type GetBalanceOfResponse struct {
+	CandidateBalance float64
+	FinalBalance     float64
+}
```
gregLibert, line 11: Too many useless information.
My suggestion:
```suggestion
type Balance struct {
	Candidate float64
	Final     float64
}
```

pkg/node/getBalanceOf.go
```diff
@@ -0,0 +1,31 @@
...
+
+// GetBalanceOf returns as a float64 the candidate balance and final balance of an address
+func GetBalanceOf(client *Client, walletAddress string) (*GetBalanceOfResponse, error) {
```
gregLibert, line 14: Names to improve:
- The balance is retreived remotely. You should clarify that by using fetch instead of Get
- The wallet part doesn't bring additional info.

My suggestion:
```suggestion
func FetchBalance(client *Client, address string) (*Balance, error) {
```

pkg/node/getBalanceOf.go
```diff
@@ -0,0 +1,31 @@
...
+	addressDetails, err := Addresses(client, []string{walletAddress})
+	if err != nil {
+		return nil, fmt.Errorf("calling get_addresses :%w", err)
```
gregLibert, line 17: If you look inside the code of Addresses, you will see that the reference to get_address is (almost, you could fix it ;-) ) already there. So this information is useless. Pass directly the error or wrap it smartly (by adding meaningful information).

pkg/node/getBalanceOf.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	candidateBalance, err := strconv.ParseFloat(addressDetails[0].CandidateBalance, 64)
+	if err != nil {
+		return nil, fmt.Errorf("converting string to float :%w", err)
```
gregLibert, line 22:
```suggestion
		return nil, fmt.Errorf("converting candidateBalance %s f64 :%w", addressDetails[0].CandidateBalance, err)
```

pkg/node/getBalanceOf.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	finalBalance, err := strconv.ParseFloat(addressDetails[0].FinalBalance, 64)
+	if err != nil {
+		return nil, fmt.Errorf("converting string to float :%w", err)
```
gregLibert, line 27:
```suggestion
		return nil, fmt.Errorf("converting candidateBalance %s f64 :%w", addressDetails[0].FinalBalance, err)
```

gregLibert (changes requested):
:no_good: Code clarity can be improved.

pkg/node/fetchBalance.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+// FetchBalance returns as a float64 the candidate balance and final balance of an address.
+func FetchBalance(client *Client, address string) (*Balance, error) {
+	addressDetails, err := Addresses(client, []string{address})
+	if err != nil {
+		return nil, err
+	}
+
+	candidate, err := strconv.ParseFloat(addressDetails[0].CandidateBalance, 64)
```
peterjah, line 20: We should not use floating points for balance & token amounts.
My proposition would be to use this library:
https://pkg.go.dev/github.com/shopspring/decimal

peterjah (changes requested):

gregLibert (approved):

pkg/node/fetchBalance.go
gregLibert: I approve that!

peterjah (approved):

gregLibert (approved):

PR #429, opened 2023-01-24
386 plugin manager front

0xMazout:


0xMazout:
Take this in mind when you review plis :
https://github.com/massalabs/thyra/issues/463
https://github.com/massalabs/thyra/issues/462

api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -121,6 +121,30 @@ paths:
...
+        - text/html
+        - text/css
+        - text/webp
+        - image/png
+      parameters:
+        - in: path
+          name: resource
+          type: string
+          default: index.html
+          required: true
```
gregLibert, line 139: Not sure you can:
- define a default value if parameter is in path
- require is true by default for parameter in path

What is the output of swagger when you ask it to validate your swagger file ?

int/api/thyra.go
```diff
@@ -78,7 +78,16 @@ func ThyraRegistryHandler(params operations.ThyraRegistryParams) middleware.Resp
...
+	content, err := contentReact.ReadFile(basePathReact + "home/" + params.Resource)
+	if err != nil {
+		return operations.NewThyraHomeNotFound()
+	}
+
+	return NewCustomResponder(content, contentType(params.Resource), http.StatusOK)
+}
+
+func ThyraPluginManagerHandler(params operations.ThyraPluginManagerParams) middleware.Responder {
+	content, err := contentReact.ReadFile(basePathReact + "plugin-manager/" + params.Resource)
```
gregLibert, line 90: This code is duplicated a lot.

web/plugin-manager/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+lerna-debug.log*
```
gregLibert, line 8:
```suggestion
*.log*
```

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+    // fetch info from plugin to get fresh data on demand
+    async function fetchPluginStatus(): Promise<AxiosResponse<string>> {
+        let result: AxiosResponse<string> = {} as AxiosResponse<string>;
+        try {
+            return (result = await axiosServices.getpluginInfo(dataMemoized.id));
+        } catch (error) {
+            sendErrorData(
+                "error",
+                `Plugins infos failed to get infos from plugin name : ${dataMemoized.name} on id: ${dataMemoized.id}`
```
gregLibert, line 32: You're not getting information from the error content ?

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
+import { SetStateAction, useMemo, useState } from "react";
+import { ArrowPathIcon, TrashIcon, PlayCircleIcon } from "@heroicons/react/24/outline";
+import axiosServices from "../services/axios";
+import { AxiosResponse } from "axios";
+import { Plugin, PluginProps, PluginStatus } from "../../../shared/interfaces/IPlugin";
+import { statusHelper } from "../helpers/statusHelpers";
+
+function PluginBlock(p: PluginProps) {
```
gregLibert, line 8: Is it clean to use a function of functions like that? Aren't you suppose to use a class ?

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+    // Launch or stop plugin
+    async function launchAndStopPlugins() {
+        let resultPluginInfo: AxiosResponse<string>;
+        // fetch info from plugin to get fresh data
+        resultPluginInfo = await fetchPluginStatus();
+        // Update data
+        setStatus(statusHelper(resultPluginInfo.data));
+
+        let result: AxiosResponse<number>;
+        if (!toggleStatus) {
```
gregLibert, line 47: why not using this.toggleStatus ? This is linked to the class question. I have the feeling that the code is unclear has you don't immeditely know where this variable come from.

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+import { Plugin, PluginProps, PluginStatus } from "../../../shared/interfaces/IPlugin";
+import { statusHelper } from "../helpers/statusHelpers";
+
+function PluginBlock(p: PluginProps) {
+    // Callback to set error on parent
+    function sendErrorData(errorType: string, errorMessage: string) {
+        p.setErrorData(errorType, errorMessage);
+    }
+    // Data to display
+    let dataMemoized = useMemo(() => {
```
gregLibert, line 14: Are you sure you want that the reader remember that "this data is memoized" or do you think it's better than the reader remember that these are plugin properties ?

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+        // Update data
+        setStatus(statusHelper(resultPluginInfo.data));
+
+        let result: AxiosResponse<number>;
+        if (!toggleStatus) {
+            // Launch plugin
+            try {
+                result = await axiosServices.manageLifePlugins(dataMemoized.id, "start");
+                // dataMemoized.status = PluginStatus.Up;
+                setStatus(!toggleStatus);
```
gregLibert, line 52: Shouldn't you use diretcly the targeted status to clarify the code ?

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+                forcePlayStatus(true)
+                return result;
+            } catch (error) {
+                sendErrorData("error", "Start plugin failed");
+                // dataMemoized.status = PluginStatus.Error;
```
gregLibert, line 57: dead code should be removed

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+        if (!toggleStatus) {
+            // Launch plugin
+            try {
+                result = await axiosServices.manageLifePlugins(dataMemoized.id, "start");
+                // dataMemoized.status = PluginStatus.Up;
+                setStatus(!toggleStatus);
+                forcePlayStatus(true)
+                return result;
+            } catch (error) {
+                sendErrorData("error", "Start plugin failed");
```
gregLibert, line 56: no information extracted on actual error ?

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+                // dataMemoized.status = PluginStatus.Error;
+                return;
+            }
+        } else {
```
gregLibert, line 60: useless else as you return in every case in the if.

web/plugin-manager/src/components/pluginBlock.tsx
gregLibert, line 47: Why not handling the passing case first ?

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+                return;
+            }
+        }
+    }
+
+    // Update plugin
+    // Not implemented atm
+    function updatePlugins() {
+        //Front end update
+        // propsRef.current.updateAvailable = !propsRef.current.updateAvailable
```
gregLibert, line 80: dead code ?

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+        //     sendErrorData("error","Plugins infos failed to launch")
+        // }
+        // propsRef.current.online = true
+        // return result;
+        //#####################################################
+        console.log("Update is Not implemented ATM");
+    }
+    // Open plugin homepage
+    function openHomepagePlugins() {
+        if (statusHelper(dataMemoized.status)) window.open(dataMemoized.home);
```
gregLibert, line 95: statusHelper name doesn't really help me understand what this function is doing, except helping :-)

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+    function removePlugins() {
+        try {
+            axiosServices.deletePlugins(dataMemoized.id);
+            p.triggerRefreshPluginList();
+            sendErrorData("success", "Plugin removed");
+        } catch (error) {
+            sendErrorData("error", "Plugins failed to be removed");
+        }
+    }
+    // Minimize string to fit in the block
```
gregLibert, line 110:
```suggestion
    // Truncate the string so that it fits in the given lenght if needed.
```

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+    function minimizeString(str: string, length: number) {
```
gregLibert, line 111: String part is useless. And minimize is unclear, you should use a clearer function name based on what is done.

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+        } else {
+            return str;
+        }
+    }
+
+    // Change the play status icon color and update the status if we want to force it
+    function playStatus() {
+            return statusHelper(dataMemoized.status)
+            ? "w-6 h-6 text-green-500"
+            : "w-6 h-6 text-red-500";
```
gregLibert, line 123: not sure [Ternary conditional operator](https://en.wikipedia.org/wiki/Ternary_conditional_operator) makes thing easier to read here.

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+    function minimizeString(str: string, length: number) {
+        if (str.length > length) {
+            return str.substring(0, length) + "...";
+        } else {
+            return str;
+        }
+    }
+
+    // Change the play status icon color and update the status if we want to force it
+    function playStatus() {
```
gregLibert, line 120: This function name is misleading : if I read it without looking at the code I have the feeling that you will play a status.

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,183 @@
...
+    function forcePlayStatus(status: boolean) {
+        if (status) {
+            setPlayStatusClassName("w-6 h-6 text-green-500");
+        } else {
+            setPlayStatusClassName("w-6 h-6 text-red-500");
+        }
+    }
+
+    // Return the right icon for the update
+    function updateStatus() {
```
gregLibert, line 135: the function name is not matching what it does.

web/plugin-manager/src/helpers/alertHelpers.tsx
```diff
@@ -0,0 +1,90 @@
+import React from "react";
+
+function alertHelper(typeAlert: string, message: string, remove: () => void) {
+    const successAlert = (
```
gregLibert, line 4: Why the const are defined in the function? This means that you will have 3 different const set each time you call the alerthelper. Don't you think this should be defined outside of the function, hence being defined only once ?

web/plugin-manager/src/helpers/alertHelpers.tsx
```diff
@@ -0,0 +1,90 @@
...
+            returnVal = successAlert;
+            break;
+        case "error":
+            returnVal = errorAlert;
+            break;
+        default:
+            returnVal = warningAlert;
+            break;
+    }
+    return returnVal;
```
gregLibert, line 87: Why using a useless variable returnVal here and directly returns in the switch below ? WHich practice is the good one ?

web/plugin-manager/src/helpers/statusHelpers.ts
```diff
@@ -0,0 +1,16 @@
+export const statusHelper = (status: string) => {
```
gregLibert, line 1: in general when you rturn a boolean you should use a isXXX or hasXXX.

web/plugin-manager/src/pages/Manager.tsx
```diff
@@ -0,0 +1,112 @@
...
+    //State to store plugins populated
+    const [pluginsPopulated, setpluginsPopulated] = useState([<PuffLoader/>])
+    //Callback to remove Error
+    function removeError(): void {
+        setError(<></>);
+    }
+    function setErrorHandler(errorType: string, errorMessage: string): void {
+        setError(alertHelper(errorType, errorMessage, removeError));
+        setInterval(() => {
+            setError(<></>);
```
gregLibert, line 23: why not calling removeError ?

web/plugin-manager/src/pages/Manager.tsx
```diff
@@ -0,0 +1,112 @@
...
+        }, 10000);
+    }
+    // Initialize UI on render
+    const initializeUi = async () => {
+        pluginsInfos = await axiosServices.getPluginsInfo();
```
gregLibert, line 28: is this variable used ? Where ? How ?

web/plugin-manager/src/pages/Manager.tsx
```diff
@@ -0,0 +1,112 @@
...
+        populatePlugins();
+    };
+
+    async function getPluginsInfo () {
+        try {
+            pluginsInfos = await axiosServices.getPluginsInfo();
+            console.log(pluginsInfos)
```
gregLibert, line 35: needed ?

web/plugin-manager/src/pages/Manager.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }
+    // Initialize UI on render
+    const initializeUi = async () => {
+        pluginsInfos = await axiosServices.getPluginsInfo();
+        populatePlugins();
+    };
+
+    async function getPluginsInfo () {
+        try {
+            pluginsInfos = await axiosServices.getPluginsInfo();
```
gregLibert, line 34: I have the feeling that it's the same thing as initializeUI.

web/plugin-manager/src/pages/Manager.tsx
```diff
@@ -0,0 +1,112 @@
...
+        setError(<></>);
+    }
+    function setErrorHandler(errorType: string, errorMessage: string): void {
+        setError(alertHelper(errorType, errorMessage, removeError));
+        setInterval(() => {
+            setError(<></>);
+        }, 10000);
+    }
+    // Initialize UI on render
+    const initializeUi = async () => {
```
gregLibert, line 27:
```suggestion
    const initializeUI = async () => {
```

web/plugin-manager/src/pages/Manager.tsx
```diff
@@ -0,0 +1,112 @@
...
+        const interval = setInterval(async () => {
+            getPluginsInfo();
+        }, 10000);
+        return () => clearInterval(interval);
+    }, []);
+
+    const mock: Plugin = {
+        name: "Plugin 1",
+        logo:
+            "https://upload.wikimedia.org/wikipedia/fr/thumb/1/15/Audi_logo.svg/1280px-Audi_logo.svg.png",
```
gregLibert, line 56: If it's a problem don't you thing that another image should be used ?

web/plugin-manager/src/pages/Manager.tsx
```diff
@@ -0,0 +1,112 @@
...
+        home: "/urlOfPlugin",
+        // isUpdate: true,
+        id: "1",
+    };
+    // Mocks in case we don't have the plugin manager
+    let mocks = [mock];
+
+    function populatePlugins () {
+        console.log(pluginsInfos.status)
+        setpluginsPopulated((pluginsInfos.status == 200)
```
gregLibert, line 69: difficult to read.

web/plugin-manager/src/pages/Manager.tsx
```diff
@@ -0,0 +1,112 @@
...
+        status: "Down",
+        home: "/urlOfPlugin",
+        // isUpdate: true,
+        id: "1",
+    };
+    // Mocks in case we don't have the plugin manager
+    let mocks = [mock];
+
+    function populatePlugins () {
+        console.log(pluginsInfos.status)
```
gregLibert, line 68: useful ?

web/plugin-manager/src/pages/Manager.tsx
```diff
@@ -0,0 +1,112 @@
...
+            ? 
+            pluginsInfos.data.map((mock: Plugin) => {
+                    let pluginProps: PluginProps = {
+                        props: mock,
+                        setErrorData: setErrorHandler,
+                        triggerRefreshPluginList: function (): void {
+                            getPluginsInfo();
+                        }
+                    };
+                    console.log(mock)
```
gregLibert, line 79: useful?

gregLibert (changes requested):

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: Yes it's normal. React Functionnal component is build for it with Hooks and Functionnal component we don't use classes. Hooks is not compatible with classes.

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: it's a state it's normal

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: No the Else is only if the status of the plugin is stoped . it's not useless . 100% sure  i tested it.

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: Disagree . Pretty clear and it's a good practice to do it

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: Because Start process first . it's opinionated.

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: image
Seems pretty clear.

It's a helper for the front, We receive string then we transform in a bool to state if we can launch or not. If you have another proposal feel free to propose.

web/plugin-manager/src/helpers/alertHelpers.tsx
0xMazout: done

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: changed for : definePlayStatus

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: changed for : defineUpdateStatus

web/plugin-manager/src/helpers/alertHelpers.tsx
0xMazout: Changed in functions to avoid the memory allocation. But it's not significant in term of gains

web/plugin-manager/src/helpers/statusHelpers.ts
0xMazout: changed for isStatusReady

web/plugin-manager/src/pages/Manager.tsx
0xMazout: because it's a setState convention react

web/plugin-manager/src/pages/Manager.tsx
0xMazout: image

web/plugin-manager/src/pages/Manager.tsx
0xMazout: dead code

web/plugin-manager/src/pages/Manager.tsx
0xMazout: yes was a late miss shot.

web/plugin-manager/src/pages/Manager.tsx
0xMazout: deleted

web/plugin-manager/src/pages/Manager.tsx
0xMazout: For me its the same. 
image

Will push the if else. But ternary are not demons.

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: No because if im in the try and result didn't fail. status was false and now he has to be true.

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: The goal was to have readable data. and to have a good ui.
If i display error.message it looks like shit

web/plugin-manager/src/components/pluginBlock.tsx
gregLibert: What is your source ? Mine is [The art of readable code](https://mcusoft.files.wordpress.com/2015/04/the-art-of-readable-code.pdf), chapter 7 Making Control Flow Easy to Read.

Compared to the following code your gaining two lines...

```typescript
if statusHelper(dataMemoized.status) {
    return "w-6 h-6 text-green-500";
} else {
    return "w-6 h-6 text-red-500";
}
```

web/plugin-manager/src/components/pluginBlock.tsx
gregLibert: The function name is not clear:
- no verb used
- it doesn't explain what the function is really doing.

Using a function name sucha as: isUp(status) or isRunning(status) would make things easier to understand.

web/plugin-manager/src/components/pluginBlock.tsx
gregLibert: My practice to start with the positive case first is extarcted from [The art of readable code](https://mcusoft.files.wordpress.com/2015/04/the-art-of-readable-code.pdf), chapter 7 The Order of if/else Blocks

web/plugin-manager/src/components/pluginBlock.tsx
gregLibert: My question is not regarding the useMemo function (that I didn't know so thank you for this information :-)). it was about the variable name : dataMemoized. You should nema it to something usefull for the code reader. Something like `pluginProperties`, `pluginAttributes` or `pluginStatus` based on what is contained in this variable.

web/plugin-manager/src/components/pluginBlock.tsx
gregLibert: Sounds good, but how can we access the error to debug if needed ? Are they logged in the console ? Should we plan a king of collapsable area to get additionnal information ? depending on the solution you image, can be done in another issue :-)

web/plugin-manager/src/components/pluginBlock.tsx
gregLibert: your answer is unclear to me. Maybe we should discuss that

0xMazout: https://www.geeksforgeeks.org/differences-between-functional-components-and-class-components-in-react/

because we are in functionnal component we don't create classes and we use Hooks.

UseState is a Hook .
toggleStatus is a state defined by UseState. it's a way to store value through renders. If you don't store it in a hook your value is lost 

To access to toggleStatus. i use the variable.

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: Done

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: Done

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: Done

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: include error.message in all alerts

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: Done

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: Done

int/api/thyra.go
0xMazout: the line is duplicated two times at 81 and 90 all the rest of the code is different.
And as you said: We won't create a function to avoid a single line . 

So i think it's unnecesary to factorize this at this moment.

api/swagger/server/restapi/resource/swagger.yml
0xMazout: Honestly it's a copy pasta from all of others thyra/**plugins**/{resources}

And it works well on them.

image

image

I didn't re-invent the wheel and i think go generate don't print any error
And swagger validate print this :

image

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: Messages was personnalized on each error. So you can find it in the code. It was pretty same has we have in Thyra, except they are not in library. 

but i changed and now we display error.message.
And we will see in the future if we want to build something else. 

But for me in any case User don't have to know if error was 400 or 404, If he want to see this he can open the console and inspect the request. But Madame Michu doesn't care about this and if she need a debug she can copy the message and send it to us.

api/swagger/server/restapi/resource/swagger.yml
0xMazout: i tried to delete the default value. and go validate still show this

0xMazout: i delete the default value. but i don't know if it breaks something else because this was not related to my subject

web/plugin-manager/src/components/pluginBlock.tsx
gregLibert: Thank you, I learned something.

int/api/thyra.go
gregLibert: @0xMazout issue created here: https://github.com/massalabs/thyra-plugin-hello-world/issues/7

web/plugin-manager/src/components/pluginBlock.tsx
gregLibert: Then I don't understand. Let's talk about this.

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,180 @@
...
+                "error",
+                `Plugins infos failed to get infos from 
+                plugin name : ${pluginProperties.name} on id: ${pluginProperties.id}, error: ${error.message}}`
+            );
+        }
+        return result;
+    }
+
+    // Launch or stop plugin
+    async function launchAndStopPlugins() {
```
gregLibert, line 40: should be launchOrStop, shouldn't it?

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,180 @@
...
+    function minimize(str: string, length: number) {
+        if (str.length > length) {
+            return str.substring(0, length) + "...";
+        } else {
+            return str;
+        }
+    }
+
+    // Change the play status icon color and update the status if we want to force it
+    function definePlayStatus() {
```
gregLibert, line 114: Is it play or  run. I dont understand this play notion. I will call you.

web/plugin-manager/src/helpers/alertHelpers.tsx
```diff
@@ -0,0 +1,82 @@
+function alertHelper(typeAlert: string, message: string, remove: () => void) {
+    function successAlert(){
+        return<div className="alert alert-success shadow-lg">
```
gregLibert, line 3:
```suggestion
        return <div className="alert alert-success shadow-lg">
```

web/plugin-manager/src/pages/Manager.tsx
gregLibert: I don't understand.

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -0,0 +1,180 @@
...
+    function sendErrorData(errorType: string, errorMessage: string) {
+        p.setErrorData(errorType, errorMessage);
+    }
+    // Data to display
+    let pluginProperties = useMemo(() => {
+        return p.props;
+    }, [p.props]);
+
+    // Toggle status state
+    const [toggleStatus, setStatus] = useState(isUp(p.props.status));
```
gregLibert, line 19: What is useState ?

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: done

web/plugin-manager/src/components/pluginBlock.tsx
0xMazout: play or run was the same but lets go to run

api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -111,7 +108,29 @@ paths:
         - in: path
           name: resource
           type: string
-          default: index.html
+          required: true
```
gregLibert, line 111: to be removed

gregLibert (approved):
:clap: Goog job!
You deserve the badge :camping:!

SebastienValla (approved):
Nice !

PR #434, opened 2023-01-27
Fix for MIME type handling

mickael-btc:
Hello,

I have fixed an issue with the MIME type handling in the server. The problem was that the server was not properly handling MIME types for certain file extensions, which caused errors when loading scripts and other resources.

For example when uploading a website that was built with react, the javascript files weren’t accepted as valid javascript modules since thyra marked them as json.

I have added common MIME types i found on [MDM](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types). I updated the setContentType function to handle the vast majority of extensions.

I have tested the changes and they are working correctly. I would like to request that these changes be included in the main branch.

Thank you for your time and consideration.

### DIFF
--- a/pkg/onchain/website/browser.go
+++ b/pkg/onchain/website/browser.go
@@ -30,18 +30,96 @@ func removeEmptyStrings(s []string) []string {
 	return result
 }
 
+// Common MIME types from https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+var contentTypeMap = map[string]string{
+	".aac":    "audio/aac",
+	".abw":    "application/x-abiword",
+	".arc":    "application/x-freearc",
+	".avi":    "video/x-msvideo",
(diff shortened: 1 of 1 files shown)


peterjah (changes requested):
Thanks for your contribution.
I think it would be better to not deal with these MIME types and directly use the mime package from go
https://pkg.go.dev/mime
it does quite the same thing as you:
```
		file := "test.html"
		ctype := mime.TypeByExtension(filepath.Ext(file))
		w.Header().Set("Content-Type", ctype)
```
The good think is you don't have to maintain this map of extensions/mime types

peterjah:
done here:
https://github.com/massalabs/thyra/pull/438

can you confirm it fits your needs !

mickael-btc:
Yes thank you !

peterjah:
merged 
https://github.com/massalabs/thyra/pull/438

PR #438, opened 2023-01-30
support all MIME types for website upload

peterjah:


Thykof (approved):

gregLibert (approved):
:thinking: the code is clean.

But don't you think that reusing a well construct and export package like for instance in the hello world can reduce the code copied/pasted?

PR #452, opened 2023-02-07
update plugin register infos

peterjah:


0xMazout (approved):
LGTM

api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -736,10 +736,17 @@ paths:
                   description: Plugin description.
                   type: string
                   x-nullable: false
+                home:
```
gregLibert, line 739: Why do you need that? Why don't you use /thyra/plugin/{author-name}/{plugin-name} as home ?

int/api/myplugin/list.go
```diff
@@ -41,9 +42,13 @@ func (l *list) Handle(param operations.PluginManagerListParams) middleware.Respo
 		info := plgn.Information()
 
 		if info != nil {
+			pluginURL := fmt.Sprintf("%s%s/%s/", plugin.EndpointPattern, url.PathEscape(info.Author), url.PathEscape(info.Name))
```
gregLibert, line 45: The missing / in '%s%s/%s/' sounds weird.

api/swagger/server/restapi/resource/swagger.yml
gregLibert: This si done to make things easier on the front part.

gregLibert (approved):

PR #453, opened 2023-02-07
fix mac os arch for py installer

peterjah:


thomas-senechal:
Hi @peterjah !

There is no issue here. Pyinstaller uses `x86_64` for `amd64` processors. See [here](https://pyinstaller.org/en/stable/feature-notes.html#macos-multi-arch-support) the list of available targets

peterjah:
> Hi @peterjah !
> 
> There is no issue here. Pyinstaller uses `x86_64` for `amd64` processors. See [here](https://pyinstaller.org/en/stable/feature-notes.html#macos-multi-arch-support) the list of available targets

Hooo nice! hopefully you checked it 😄 
I close the issue!

PR #455, opened 2023-02-08
435 durable and resilient thyra install script

thomas-senechal:
Latest artifacts: https://github.com/massalabs/thyra/actions/runs/4126473454


scripts/installer.py
```diff
@@ -0,0 +1,161 @@
...
+        logging.info(filename + " downloaded successfully")
+
+    """
+    This method must be implemented by the installer of each platform to install and configure the DNS server.
+    """
+    def setupDNS(self):
+        pass
+
+    """
+    Generates an HTTPS certificate for my.massa using mkcert and stores it in the thyra config folder.
```
gregLibert, line 91: The certificate is the CA and not the certificate for my.massa, isn't it ?

scripts/installer.py
```diff
@@ -0,0 +1,161 @@
...
+            self.printErrorAndExit(f"Error while deleting mkcert binary: {err}")
+        logging.info("HTTPS certificate successfully generated")
+
+    """
+    Downloads thyra server and stores it in the thyra install folder.
+    """
+    def installThyraServer(self):
+        self.downloadFile(self.THYRA_SERVER_URL, self.THYRA_SERVER_FILENAME)
+        os.chmod(self.THYRA_SERVER_FILENAME, 0o755)
+        if os.getcwd() != self.THYRA_INSTALL_FOLDER_PATH:
```
gregLibert, line 126: And if not you install nothing  without any log ?
Why are you doing this test ?

scripts/installer.py
```diff
@@ -0,0 +1,161 @@
...
+        os.chmod(self.THYRA_APP_FILENAME, 0o755)
+        if os.getcwd() != self.THYRA_INSTALL_FOLDER_PATH:
+            try:
+                thyra_app_path = os.path.join(self.THYRA_INSTALL_FOLDER_PATH, self.THYRA_APP_FILENAME)
+                if os.path.exists(thyra_app_path):
+                    os.remove(thyra_app_path)
+                shutil.move(self.THYRA_APP_FILENAME, self.THYRA_INSTALL_FOLDER_PATH)
+            except OSError as err:
+                self.printErrorAndExit(f"Error while moving thyra app binary: {err}")
+        logging.info("Thyra app installed successfully")
```
gregLibert, line 150: looks like a duplication of the previous function.

scripts/installer.py
```diff
@@ -0,0 +1,161 @@
...
+
+    """
+    Installs thyra server, thyra app and a DNS server.
+    """
+    def startInstall(self):
+        logging.info("Starting installation of thyra")
+        self.installThyraServer()
+        self.installThyraApp()
+        self.setupDNS()
```
gregLibert, line 159: You should setup and configure the DNs only if my.massa is not resolved.

scripts/installer.py
```diff
@@ -0,0 +1,161 @@
...
+        self.generateCertificate()
```
gregLibert, line 160: You shouldgenerate certificate only if it doesn't already exist.

scripts/installer.py
```diff
@@ -0,0 +1,161 @@
...
+            except OSError as err:
+                self.printErrorAndExit(f"Error while creating certs folder: {err}")
+
+        self.downloadFile(self.MKCERT_URL, self.MKCERT_FILENAME)
+        os.chmod(self.MKCERT_FILENAME, 0o755)
+
+        self.executeCommand([
+            os.path.join(os.getcwd(), self.MKCERT_FILENAME), 
+            "--install"])
+        self.executeCommand([
```
gregLibert, line 108: By experience mkcert is very verbose but share some information on compatibility or action to perform after installation, such as restart the browser for instance. By using this command, will the user get all those messages?

scripts/windows_installer.py
```diff
@@ -0,0 +1,104 @@
...
+                os.remove(self.ACRYLIC_DNS_PROXY_FILENAME)
+            except:
+                logging.error("Could not remove Acrylic DNS Proxy installer")
+
+        self.configureAcrylic()
+
+
+if __name__ == "__main__":
+    if platform.system() != "Windows":
+        logging.error("This script is only compatible with Windows")
```
ASAPSegfault, line 94: Why don't you use the helper function printErrorAndExit() ?

scripts/windows_installer.py
```diff
@@ -0,0 +1,104 @@
...
+        os.system("pause")
+        os._exit(-1)
+
+    if ctypes.windll.shell32.IsUserAnAdmin() == 0:
+        logging.error("This script must be run as administrator")
```
ASAPSegfault, line 99: Same here

scripts/macos_installer.py
```diff
@@ -0,0 +1,55 @@
...
+
+        self.executeCommand("sudo bash -c 'echo ""address=/.massa/127.0.0.1"" > $(brew --prefix)/etc/dnsmasq.d/massa.conf'", True)
+        self.executeCommand("sudo mkdir -p /etc/resolver", True)
+        self.executeCommand("sudo bash -c 'echo ""nameserver 127.0.0.1"" > /etc/resolver/massa'", True)
+
+        logging.info("Restarting DNSMasq...")
+        self.executeCommand("sudo brew services restart dnsmasq", True)
+
+    def setupDNS(self):
+        runningDNS = self.executeCommand("sudo lsof -i :53 | sed -n 2p | sed 's/[[:space:]].*$//'", True)
```
gregLibert, line 36: This is pure shell. Can't you do that in a more python idiomatic way ?

scripts/macos_installer.py
```diff
@@ -0,0 +1,55 @@
...
+            self.MKCERT_URL = "https://dl.filippo.io/mkcert/latest?for=darwin/amd64"
+        else:
+            self.printErrorAndExit(f"Unsupported architecture {platform.machine()}")
+
+    def configureDNSMasq(self):
+        logging.info("Configuring DNSMasq...")
+
+        self.executeCommand("sudo bash -c 'echo ""address=/.massa/127.0.0.1"" > $(brew --prefix)/etc/dnsmasq.d/massa.conf'", True)
+        self.executeCommand("sudo mkdir -p /etc/resolver", True)
+        self.executeCommand("sudo bash -c 'echo ""nameserver 127.0.0.1"" > /etc/resolver/massa'", True)
```
gregLibert, line 30: This is pure shell. can't you do that in a more python idiomatic way ?

scripts/windows_installer.py
```diff
@@ -0,0 +1,104 @@
...
+        logging.info("Configuring network interface...")
+        commandOutput = self.executeCommand("wmic nic where \"netenabled=true\" get netconnectionID", True)
+
+        networkAdpatersNames = list(filter(None, commandOutput.split('\n')))
+        networkAdpatersNames.pop(0)
+        networkAdpatersNames = [name.strip() for name in networkAdpatersNames]
+
+        for name in networkAdpatersNames:
+            self.executeCommand(f'NETSH interface ipv4 set dnsservers "{name}" static 127.0.0.1 primary')
+            self.executeCommand(f'NETSH interface ipv6 set dnsservers "{name}" static ::1 primary')
```
gregLibert, line 63: Can't you do that in a more python idiomatic way ?

ASAPSegfault (changes requested):

scripts/installer.py
thomas-senechal: If not, we don't need to do what's inside of the if since we already installed Thyra in the right place.

scripts/installer.py
thomas-senechal: Yes. I thought about making a `installBinary` function that would take, an URL, a binary name and an installation path that would do the same but I didn't did it thinking the Windows install might require some specific things. 
I will add a function to simplify things.

scripts/installer.py
thomas-senechal: In fact, it's what I do in the overwritten `setupDNS` functions for [MacOS ](https://github.com/massalabs/thyra/blob/c707eeda05627cf52deda47329d291441b4b7e37/scripts/macos_installer.py#L35) and [Windows](https://github.com/massalabs/thyra/blob/c707eeda05627cf52deda47329d291441b4b7e37/scripts/windows_installer.py#L67).

Do you think I should add a `isDNSInstalled` function in the Installer, that must be overwritten in his child classes ? That function would return yes or no depending on if we need to install the DNS or not ?

scripts/installer.py
thomas-senechal: No. But `executeCommand` returns the standard output, so we can display it.

scripts/windows_installer.py
thomas-senechal: For now, the `printErrorAndExit` is in the Installer class, that's why I'm not using it there. 
I will find a clean way to call it while making sure it's still compatible with every platforms. (I will probably just make it as a static method)

scripts/installer.py
gregLibert: Thyra is installed, but is it the last version ?

scripts/installer.py
gregLibert: Yes please. The class will be easier to read.

scripts/installer.py
gregLibert: This is a good question. I think we should have the log:
- in the terminal
- in a dedicated file that can be shared in cased of installation failure

thomas-senechal: I already made some changes related to logs.
Right now, we have logs in the terminal and in a dedicated file `thyra_installer.log`. Those logs are exactly the same.
See https://github.com/massalabs/thyra/blob/c707eeda05627cf52deda47329d291441b4b7e37/scripts/installer.py#L11

About the output of mkcert Install, since `executeCommand` returns the standard output, I can print the result of the function on the standard output (not in the logs) so the user will be noticed about what happened. Or, I can print it using the logger so the user can view it in the installer logs if needed.

scripts/installer.py
thomas-senechal: I can add a check for that by running thyra using the `--version`.

gregLibert: You can. You can also systematically reinstall thyra if it's done faster as this is not explicitly ask for this task.

@SlnPons wdyt?

SlnPons: @gregLibert @thomas-senechal the best action for the user will always be the fastest when it comes to installing sth. As long as it is cleaned which means that we don't install twice or take more space that we should. 
Does it help @thomas-senechal ?

thomas-senechal: I think we should keep it as it is, so, reinstalling Thyra at each run. Users might restart the installer if something failed or to update Thyra. 

If we add a check, we should either:
- add an option to force reinstall => isn't a good idea since the purpose of the installer is to be used without a terminal.  
- ask the user if he wants to perform a reinstallation or Thyra, and then, of Thyra App (icon tray). But that would also require us to check for the latest version of Thyra locally and on Github.

I think for now, we should keep it as it is and reinstallation every time the script is run. 

WDYT @gregLibert @SlnPons ?

scripts/windows_installer.py
thomas-senechal: Couldn't find a library that would allow us to perform those actions.

scripts/installer.py
```diff
@@ -0,0 +1,195 @@
...
+    This method must be implemented by the installer of each platform to install and configure the DNS server.
+    """
+    def setupDNS(self):
+        pass
+
+    """
+    Checks if the HTTPS certificate for my.massa are already present in the thyra config folder.
+    The filenames of the certificate and the key are defined in the constants CERTIFICATION_FILENAME and CERTIFICATION_KEY_FILENAME.
+    """
+    def areCertificatesInstalled(self) -> bool:
```
gregLibert, line 109:
```suggestion
    def isCACertificateInstalled(self) -> bool:
```

scripts/installer.py
```diff
@@ -0,0 +1,195 @@
...
+        return os.path.exists(os.path.join(self.CERTIFICATIONS_FOLDER_PATH, self.CERTIFICATION_FILENAME)) and os.path.exists(os.path.join(self.CERTIFICATIONS_FOLDER_PATH, self.CERTIFICATION_KEY_FILENAME))
+
+    """
+    Installs a local Certificate Authority and generates a HTTPS certificate for my.massa.
+    """
+    def generateCertificate(self):
```
gregLibert, line 115:
```suggestion
    def generateCACertificate(self):
```

scripts/installer.py
```diff
@@ -0,0 +1,195 @@
...
+        logging.info("Generating HTTPS certificate for my.massa")
```
gregLibert, line 116:
```suggestion
        logging.info("Generating CA certificate and adding it to browsers' CA list")
```

scripts/installer.py
```diff
@@ -0,0 +1,195 @@
...
+            "--key-file", os.path.join(self.CERTIFICATIONS_FOLDER_PATH, self.CERTIFICATION_KEY_FILENAME),
+            "my.massa"])
+        if stderr is not None and len(stderr) > 0:
+            logging.info(stderr)
+
+        try:
+            os.remove(self.MKCERT_FILENAME)
+        except OSError as err:
+            self.printErrorAndExit(f"Error while deleting mkcert binary: {err}")
+        logging.info("HTTPS certificate successfully generated")
```
gregLibert, line 145:
```suggestion
        logging.info("CA certificate successfully generated")
```

scripts/installer.py
```diff
@@ -0,0 +1,195 @@
...
+        self.installThyraServer()
+        self.installThyraApp()
+        if self.shouldInstallDNS():
+            self.setupDNS()
+        else:
+            logging.info("DNS server already installed, skipping...")
+        if not self.areCertificatesInstalled():
+            self.generateCertificate()
+        else:
+            logging.info("HTTPS certificate already installed, skipping...")
```
gregLibert, line 194:
```suggestion
            logging.info("CA certificate already installed.")
```

scripts/installer.py
```diff
@@ -0,0 +1,195 @@
...
+    Installs thyra server, thyra app and a DNS server.
+    """
+    def startInstall(self):
+        logging.info("Starting installation of thyra")
+        self.installThyraServer()
+        self.installThyraApp()
+        if self.shouldInstallDNS():
+            self.setupDNS()
+        else:
+            logging.info("DNS server already installed, skipping...")
```
gregLibert, line 190:
```suggestion
            logging.info("DNS server already installed.")
```

gregLibert (changes requested):
:mag: just improving readibility. But it's really cool! Good job!

gregLibert (approved):
:clap:

ASAPSegfault (approved):

PR #456, opened 2023-02-08
Frontend ci

peterjah:


.github/workflows/tests.yml
```diff
@@ -1,7 +1,5 @@
 on:
   push:
-    branches: [ main ]
-  pull_request:
```
Thykof, line 4: do you want to keep that or is it just for you CI tests?

Thykof (approved):

SebastienValla (commented):
LGTM !

gen-frontend.go
```diff
@@ -0,0 +1,5 @@
+package resource
```
gregLibert, line 1: why not put all that in gen.go at the root of the project?

.github/actions/install/action.yml
```diff
@@ -14,6 +13,18 @@ runs:
...
+    - name: Install Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: 18
+        cache: 'npm'
+        cache-dependency-path: web/thyra-frontend/package-lock.json
+
+    - name: Build react frontend
+      shell: bash
+      run: go generate gen-frontend.go
```
gregLibert, line 30: why not regenerating everything `go generate ./...` ?

gregLibert (commented):
:question:

.github/workflows/tests.yml
peterjah: yes i want to keep! 😄 
the previous behaviour is to execute test & lint only at PR creation or on push on main.
But i think its better to execute it at every push on every branch

gen-frontend.go
peterjah: issue created:
https://github.com/massalabs/thyra/issues/461

PR #468, opened 2023-02-13
adds API continuous testing

gregLibert:


peterjah:
If we are going to use postman like this, maybe its would be nice to have a paid plan for the team to benefit from all the features (workspace sharing for example)

Also this action is deprecated because postman can now generate github action workflow
https://learning.postman.com/docs/integrations/available-integrations/ci-integrations/github-actions/#configuring-the-postman-cli-for-github-actions

It will allow us to not copy past the postman json file, because i think its gonna be a pain to maintain

peterjah (approved):

gregLibert:
> I am just wondering on which criteria theses tests should be marked as failed. right now you have endpoints returning http 400 error and test are successfull

This is driven by postman itself. If you have a look at the collection (once loaded in postman app, the json is too verbose), you will see that each endpoint has a test defined. In this test you can check whatever you want.

The endpoint having a 400 HTTP response code has no test, hence it doesn't block the CI :-).

Now, I didn't create a test for this endpoint because I need a plugin that does nothing (no registration) as this endpoint is about... registration ;-). I guess I will leave this test for a new task.

gregLibert:
> It will allow us to not copy past the postman json file, because i think its gonna be a pain to maintain

I'm not sure to understand: there is copy/paste, the collection is always with the test...

Thykof (commented):
Seen

ASAPSegfault (approved):
LGTM
It's great, we could extend this and create a Thyra collection containing all endpoints (not just the plugin manager) along with their tests.
Also it may make sense to go a bit further than just checking the response code in the test.
Like for example checking that the plugin has really been Started / Stopped / Restarted by doing a get request after checking the first request returned a 200.
It would create dependencies in the tests but at the same time from what I understand here our tests are already tied to the 'List plugin' endpoint which sets the plugin ID as a global variable, so we could use the same pattern for other endpoints.

SlnPons:
@ASAPSegfault I love what I read!! Feel free to create issues to capture your ideas. 
Not saying we're prioritise them soon but at least we have them somewhere, it's useful. 

@gregLibert how can I test this ticket on my side please?

SlnPons:
> LGTM It's great, we could extend this and create a Thyra collection containing all endpoints (not just the plugin manager) along with their tests. Also it may make sense to go a bit further than just checking the response code in the test. Like for example checking that the plugin has really been Started / Stopped / Restarted by doing a get request after checking the first request returned a 200. It would create dependencies in the tests but at the same time from what I understand here our tests are already tied to the 'List plugin' endpoint which sets the plugin ID as a global variable, so we could use the same pattern for other endpoints.

@thomas-senechal to continue what we discussed. I think parts of Hugo's suggestion would be good to improve Thyra basic testing. 
I let you have a think about the best strategy to tackle the endpoint tests and let's discuss together ;) cheers

PR #486, opened 2023-02-21
fix frontend generation

peterjah:


0xMazout (approved):

gen-frontend.go
```diff
@@ -1,5 +1,6 @@
 package resource
 
 // Thyra react frontend
-//go:generate npm install --prefix web/thyra-frontend
-//go:generate npm run --prefix web/thyra-frontend build
+//go:generate go generate web/thyra-frontend/gen.go
```
gregLibert, line 4: I think that if you do a classic `go generate ./...` you will generate the front twice. What about removing this file and add the `go generate ./...` command to the readme file ?

gregLibert (changes requested):

gen-frontend.go
peterjah: i added `go generate ./web/...` in the doc to avoid side effect by generating everything

gregLibert (approved):

PR #489, opened 2023-02-23
487 missing architecture for macos installer in releases

thomas-senechal:


.github/workflows/build.yml
```diff
@@ -104,14 +107,14 @@ jobs:
...
+          pyinstaller --clean --onefile -y --target-arch ${{ matrix.arch }} --dist ./dist/${{ matrix.target }} --workpath /tmp ${{ matrix.script }} --name thyra-installer_${{ matrix.target }}_${{ matrix.displayed_arch }}
       - name: Upload installer artifacts
         uses: actions/upload-artifact@v3
         with:
           name: thyra-installer-${{ matrix.target }}-${{ matrix.arch }}-binary
           path: |
-            ./dist/${{ matrix.target }}/macos_installer
-            ./dist/${{ matrix.target }}/windows_installer.exe
+            ./dist/${{ matrix.target }}/thyra-installer_${{ matrix.target }}_${{ matrix.displayed_arch }}
+            ./dist/${{ matrix.target }}/thyra-installer_${{ matrix.target }}_${{ matrix.displayed_arch }}.exe
```
peterjah, line 117: i doubt of what you are doing here. mixing windows and macos here seems weird.
What i usually do is set the extension in the matrix to be used like this `${{ matrix.ext }}` it return .exe for windows target and empty string for the others
see an example here:
https://github.com/massalabs/thyra-node-manager-plugin/blob/main/.github/workflows/build.yml

.github/workflows/build.yml
```diff
@@ -121,5 +124,5 @@ jobs:
           prerelease: ${{ inputs.prerelease }}
           generate_release_notes: ${{ inputs.generate_release_notes }}
           files: |
-            ./dist/${{ matrix.target }}/macos_installer
-            ./dist/${{ matrix.target }}/windows_installer.exe
+            ./dist/${{ matrix.target }}/thyra-installer_${{ matrix.target }}_${{ matrix.displayed_arch }}
+            ./dist/${{ matrix.target }}/thyra-installer_${{ matrix.target }}_${{ matrix.displayed_arch }}.exe
```
peterjah, line 128: same thing, I would be surprised if it works like this !!

peterjah (changes requested):

.github/workflows/build.yml
peterjah: every step should handle only 1 file corresponding to 1 target

.github/workflows/build.yml
thomas-senechal: Hi ! 
As you can see in [this action](https://github.com/massalabs/thyra/actions/runs/4254496743), it's working fine.
Those actions will store all the files that can be found matching the specified paths. It just throws a warning if they couldn't find any matching file. 

I searched in the Actions documentation, but they doesn't talk about this behavior. So I will update it as you suggested just in case they "fix" this behavior in an update.

peterjah: great

Thykof (approved):

PR #490, opened 2023-02-23
Remove old plugin manager

Thykof:
Remove old plugin manager and temporary code to install node manager plugin.


Thykof:
@SlnPons this PR starts the work of cleaning up thyra about plugin manager, if we merge it, old plugins that work with the legacy plugin manager won't work anymore.

I have identified:
- playground plugin
- wallet plugin: WIP https://github.com/massalabs/thyra-plugin-wallet/pull/37

peterjah (changes requested):
I think you forgot to delete the old Get plugin endpoint un the swagger yml

peterjah (approved):

gregLibert:
@Thykof  If we merge that, are you sure that the wallet plugin will still continue to work ? 
+ there is a conflict to solve

SlnPons:
@gregLibert @Thykof before merging, we need to: 
1. make sure the wallet plugin will work
2. communicate with partners to make sure they don't have important demo or other things planned with Thyra soon 
3. because we also need to install the wallet plugin by default (ie: update & clean Thyra installation scripts)

Thykof:
in order to not break anything, we must merge this after https://github.com/massalabs/thyra-plugin-wallet/pull/37 and work jointly with the other legacy plugins

peterjah (approved):

thomas-senechal:
Looks good to me, could you just rebase on main before merging ?

PR #494, opened 2023-02-28
474 improve thyra installation

thomas-senechal:


scripts/installer.py
```diff
@@ -177,6 +177,13 @@ def installThyraApp(self):
         self.installBinary(self.THYRA_INSTALL_FOLDER_PATH, self.THYRA_APP_URL, self.THYRA_APP_FILENAME)
         logging.info("Thyra app installed successfully")
 
+    def startThyraApp(self):
+        THYRA_APP_PATH = os.path.join(self.THYRA_INSTALL_FOLDER_PATH, self.THYRA_APP_FILENAME)
```
Thykof, line 181: no need to have upper case for this variable name because it is not a constant.

Thykof (changes requested):

scripts/installer.py
```diff
@@ -192,4 +199,10 @@ def startInstall(self):
             self.generateCACertificate()
         else:
             logging.info("CA certificate already installed.")
-        logging.info("Thyra installed successfully")
+
+        logging.info("Thyra installed successfully !")
+        logging.info(f"Thyra has been installed in {self.THYRA_INSTALL_FOLDER_PATH}")
```
gregLibert, line 204: From the QA needs: "Check that it is installing thyra-server and thyra-app in the directory displayed at the end of the script."

```suggestion
        logging.info(f"thyra-server and thyra-app have been installed in {self.THYRA_INSTALL_FOLDER_PATH}")
```

Furthermore, the binary name of thyra on macos is thyra_server

gregLibert (approved):

Thykof (approved):
LGTM

PR #500, opened 2023-03-01
change addresses to new format

SebastienValla:


pkg/node/sendoperation/transaction/transaction.go
```diff
@@ -34,7 +34,7 @@ func New(recipientAddress []byte, amount uint64) *Transaction {
 func (t *Transaction) Content() interface{} {
 	return &Operation{
 		Transaction: OperationDetails{
-			RecipientAddress: "A" + base58.CheckEncode(append(make([]byte, 1), t.recipientAddress...)),
+			RecipientAddress: "AU" + base58.CheckEncode(append(make([]byte, 1), t.recipientAddress...)),
```
peterjah, line 37: `"AU" + base58.VersionedCheckEncode(t.recipientAddress..., 0)`

peterjah (changes requested):
I think all conversion of address  from/to  bytes and string must be factorized.
maybe in an Address Object that will have fromPublicKey and fromPrivateKey methods

peterjah (approved):

pkg/node/sendoperation/callsc/callsc.go
```diff
@@ -33,18 +33,23 @@ type CallSC struct {
 
 func New(address []byte, function string, parameters []byte, gazLimit uint64, coins uint64,
 ) *CallSC {
+	// New testnet20 addresses needs a byte 0 for AU addresses and byte 1 for AS addresses
+	addressVersioned := append([]byte{byte(1)}, address...)
```
ASAPSegfault, line 37: Should be versionedAddress to stay consistent with the rest (b58VersionedCheckEncode for example)

ASAPSegfault (changes requested):

ASAPSegfault (approved):

PR #506, opened 2023-03-03
remove full install from lint workflow

peterjah:
To speedUp the CI, Only do the minimum for the linting process to run.

This save around 45 seconds !


Thykof (approved):
If I understand well it's a small optimization that skips the installation and only run our static code analysis, great!

SlnPons:
@Thykof have you tested this issue? 
I'm not sure it's an issue that I can test.... if yes, could you tell me how to test it please?

Issue #498, opened 2023-03-01
Https by default for Thyra pages not working


SlnPons:
**The problem(s)**
Thyra's home page and plugin-manager pages aren't in HTTPS by default. 

**To Reproduce**
1. Use case 1: 
* Go to my.massa/thyra/home/ 
* It opens http://my.massa/thyra/home  --> it should open https://my.massa/thyra/home/ by default 
* 
2. Use case 2:
* Go to https://my.massa/thyra/home/ 
* It opens http://my.massa/thyra/home/ 
* Add again https:// at the beginning of the address, it opens secure page. --> it should work the first time.

**Expected behavior(s)**
Https should be by default for all Thyra's pages.

**Screenshots**
See video: https://drive.google.com/file/d/19CrASaa8f3Hc4pQliiT_V67YIfqaZR-r/view?usp=share_link

**Thyra version**
latest as of today

**Desktop (please complete the following information):**
 - OS: MacOS
 - Google Chrome

0xMazout:
The problem comes from the interceptor in the thyra backend. 

I created a ticket about this but it was closed.

We have an interceptor in the thyra backend. each website composed by my.massa will be prefixed with http. we need to change it to https

SlnPons:
@0xMazout could you add some link to the doc about this "interceptor" please? 
If the run team works on it, it will be hard for them to get into the problem. Also is it specific to MacOS?

0xMazout:
> @0xMazout could you add some link to the doc about this "interceptor" please? If the run team works on it, it will be hard for them to get into the problem. Also is it specific to MacOS?

Yes no problem i was looking for some info on it.

0xMazout:
No it's not specific to MacOS it's related to all websites which include my.massa in their url.

In the Thyra Backend we have interceptors, to redirect on the right routes and to have format in urls. 
The sum of this interceptors give an http redirect. 

You need to isolate the specific interceptors and change it to have an https instead of an http. 

Actually i don't know what is the good interceptor.

What i can give to you is the route 
![image](https://user-images.githubusercontent.com/10938862/223150886-08621159-4aec-4e5c-b549-5239aaeff320.png)

And potential files to search 
![image](https://user-images.githubusercontent.com/10938862/223150515-a4fb1674-3b38-4707-8071-c6df03d99a87.png)


I don't have a clear solution but i think this will help you in your search

SlnPons:
Thanks a lot @0xMazout
[...]

PR #534, opened 2023-03-13
498 https by default for thyra pages not working 1

thomas-senechal:


pkg/onchain/website/handler.go
```diff
@@ -38,6 +38,16 @@ func MassaTLDInterceptor(req *interceptor.Interceptor) *interceptor.Interceptor
 	massaIndex := strings.Index(req.Request.Host, ".massa")
 
 	if massaIndex > 0 && !strings.HasPrefix(req.Request.Host, "my.massa") {
+		// If the request is not https, redirect to https.
+		if req.Request.TLS == nil {
+			http.Redirect(
+				req.Writer,
+				req.Request,
+				"https://"+req.Request.Host+req.Request.URL.Path,
```
Thykof, line 46: here we can lost query params and anchor, can't you get the full URL and replace http:// by https://?

Thykof (changes requested):

pkg/onchain/website/handler.go
thomas-senechal: > here we can lost query params and anchor
Good catch ! I will check if it's working well

> can't you get the full URL and replace http:// by https://?
No, I only have the `my.massa/thyra/...` for example. Not the protocol or other things like that.

thomas-senechal: The empty `RawQuery` was due to my WebBrowser cache. 

It should be fixed with [cd4c87c](https://github.com/massalabs/thyra/pull/534/commits/cd4c87c7a4010ebc0bd058846e0c539558d3a40b)

mazmassa (approved):
LGTM

gregLibert:
Considering my comment here : https://github.com/massalabs/thyra/issues/498#issuecomment-1460489257 I don't see why there is code for this issue : everything should be done by the browser right now.

@thomas-senechal @SlnPons we can have a quick discussion on thatif needed.

thomas-senechal:
> Considering my comment here : [#498 (comment)](https://github.com/massalabs/thyra/issues/498#issuecomment-1460489257) I don't see why there is code for this issue : everything should be done by the browser right now.
> 
> @thomas-senechal @SlnPons we can have a quick discussion on thatif needed.

Hey @gregLibert ,
As explained in [this comment](https://github.com/massalabs/thyra/issues/498#issuecomment-1461548498), the web browser does not use https by default. It depends on the web browser, its settings, its cache....

gregLibert:
> Hey @gregLibert , As explained in [this comment](https://github.com/massalabs/thyra/issues/498#issuecomment-1461548498), the web browser does not use https by default. It depends on the web browser, its settings, its cache....

Ok, but when I see:
- https://blog.mozilla.org/security/2021/08/10/firefox-91-introduces-https-by-default-in-private-browsing/#:~:text=How%20HTTPS%20by%20Default%20works,handles%20insecure%20web%20page%20addresses.
- https://blog.chromium.org/2021/03/a-safer-default-for-navigation-https.html?m=1#:~:text=Starting%20in%20version%2090%2C%20Chrome's,%E2%80%9D%20or%20%E2%80%9Chttps%3A%2F%2F%E2%80%9D.

I just hope that this is indeed a priority. Let's talk about that when we have time, no pressure.

pkg/onchain/website/handler.go
```diff
@@ -29,6 +29,26 @@ func handleMassaDomainRequest(writer http.ResponseWriter, reader *http.Request,
 	Request(writer, reader, rpcClient, addr, target)
 }
 
+// Redirects a given request to https.
+func redirectTLDToHTTPS(req *interceptor.Interceptor) {
```
gregLibert, line 33: Why not using the same approach as here: https://github.com/massalabs/thyra-plugin-hello-world/blob/main/web/endpoints.go#L33 ?

Something like:

```go
openapi.NewCustomResponder(nil, map[string]string{"Location": "https://" + r.Host + r.RequestURI}, http.StatusPermanentRedirect)
```

pkg/onchain/website/handler.go
```diff
@@ -38,6 +58,11 @@ func MassaTLDInterceptor(req *interceptor.Interceptor) *interceptor.Interceptor
 	massaIndex := strings.Index(req.Request.Host, ".massa")
 
 	if massaIndex > 0 && !strings.HasPrefix(req.Request.Host, "my.massa") {
+		// If the request is not https, redirect to https.
+		if req.Request.TLS == nil {
```
gregLibert, line 62: if you do that I think that Thyra won't work anymore for firefox on windows as HTTPS is not working for that configuration right now.

gregLibert: Furthermore, if you want to handle most requests, you should move this interceptor to the top.

pkg/onchain/website/handler.go
```diff
@@ -61,10 +86,17 @@ func RedirectToDefaultResourceInterceptor(req *interceptor.Interceptor) *interce
 		splited := removeEmptyStrings(strings.Split(req.Request.URL.Path[len(prefix):], "/"))
 
 		if len(splited) == 1 {
+			protocol := "https"
```
gregLibert, line 89: if the interceptor is moved to the top, this code is useless as the protocole here will be HTTPS when possible.

gregLibert (changes requested):

thomas-senechal:
It has been decided to "simplify" http/https handling. We wont force users to use https for `*.massa` websites.

gregLibert (approved):

PR #543, opened 2023-03-16
stop plugin on windows

SebastienValla:


Thykof (approved):

pkg/plugin/plugin.go
```diff
@@ -141,7 +141,7 @@ func (p *Plugin) Start() error {
 	// start a goroutine to wait on the command
 	go func() {
 		err := p.command.Wait()
-		if err != nil && err.Error() != "signal: killed" {
+		if err != nil && !(err.Error() == "signal: killed" || err.Error() == "exit status 1") {
```
thomas-senechal, line 144: Are we sure it's always exactly "exit status 1" ?
Shouldn't we try to check if the error contains "exit status" to make sure to catch any possible exit status such as `-1`, `1`... ?

thomas-senechal (approved):

PR #546, opened 2023-03-16
auto mime type

SebastienValla:


int/api/utils.go
```diff
@@ -1,29 +1,26 @@
...
-		contentType = map[string]string{"Content-Type": "text/javascript"}
-	case ".html":
-		contentType = map[string]string{"Content-Type": "text/html"}
-	case ".webp":
-		contentType = map[string]string{"Content-Type": "text/webp"}
-	case ".png":
-		contentType = map[string]string{"Content-Type": "image/png"}
-	default:
-		contentType = map[string]string{}
+	if filepath.Ext(rsc) == ".otf" {
```
ASAPSegfault, line 15: You should create a variable to store `filepath.Ext(rsc)`.
It will be clearer if we have other edge cases

ASAPSegfault (changes requested):

Thykof (approved):
unfortunatly otf is not handled by "mime" https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/mime/type.go;l=60

ASAPSegfault (approved):

Issue #559, opened 2023-03-27
Add `cmd` endpoint tests


thomas-senechal:
**Context**
Following https://github.com/massalabs/thyra/issues/514 , we need to test every parts of Thyra endpoints using the Robot Framework. The task has been divided in sub issues to allow parallelisation of endpoint testing.

**How to**

- [x] Write tests endpoint by endpoint
- [x] Write scenarios tests if possible

**Technical details**
- Create a `cmd.robot` test file
- Test all `cmd` endpoints
- Write a test that will upload a SC and execute a function of that SC
- Make sure that no important informations have been missed from https://github.com/massalabs/thyra/issues/504
- [x] Following [this comment](https://github.com/massalabs/thyra/pull/564#discussion_r1153153832):
    - Remove the `.wasm` file.
    - Use the npm js package @massalabs/sc-standards to build the DNS and upload the .wasm file.

thomas-senechal:
Since those endpoints needs a wallet with some coins in it, we had a meeting with Gregory to know how to have a Wallet with enough coins for our tests. We were planning to use a Private Key in the Github Secrets to allow us to access the Wallet and so, sign transactions. 

The issue is that the Plugin Wallet (and Thyra wallet) are using a GUI to import wallets and sign transactions. 

A way to solve the issue with wallet import is to create a wallet and store the file in the Github Secrets. This would allow us to bypass the GUI for importing a wallet, but the issue remains for signing transactions.

As we do in Thyra e2e tests, I was planning to use `SikuliX` to be able to fill the password in the GUI to sign transactions. There is a Robot Framework library for `SikuliX` that could allow us to use `SikuliX` from Robot Framework tests. I tried to install it but never succeeded at starting the `SikuliX` library. It looks like the setup process is a bit messy.

thomas-senechal:
If we can't fix this GUI issue for signing transactions, we won't be able to tests the following endpoints:
- `cmd` endpoints (this issue)
- Web on Chain: We need to be able to sign transactions to upload a website. If we can't upload a website, we won't be able to check if we can retrieve a website, or the list of websites...
- Event Manager: To test the event manager, we need a way to generate events to make sure we are catching the right events. A way of doing this is to deploy a SC that can generate events and to execute a function in it and check if we get the generated event. But since we can't sign transaction, we can't deploy nor execute a function.

Issue #514, opened 2023-03-07
Test Thyra endpoints


thomas-senechal:
**Context**
Thyra is mostly stable and must be tested to ensure that every features are working as expected. For that, we need to test Thyra's endpoints.

**User flow**
No user flow since it's a purely development oriented task

**How to**

- [x] List the parts of Thyra that needs to be tested
- [ ] Make sure all endpoints are well tested
    - [x] https://github.com/massalabs/thyra/issues/559
    - [x] https://github.com/massalabs/thyra/issues/560
    - [x] https://github.com/massalabs/thyra/issues/561
    - [x] https://github.com/massalabs/thyra/issues/562
    - [x] https://github.com/massalabs/thyra/issues/585
    - [ ] WebOnChain: 
        - [x] https://github.com/massalabs/thyra/issues/600
        - [ ] https://github.com/massalabs/thyra/issues/601
        - [x] https://github.com/massalabs/thyra/issues/613
        - [ ] https://github.com/massalabs/thyra/issues/614
- [ ] Make sure to have some "scenarios" that will match how we use the endpoints in Thyra

**Technical details**

- Create a `.robot` test file for each parts that needs to be tested
- Write tests for each part of Thyra in the corresponding `.robot` test file
- Once every related issues are done:
    - [ ] deleted KPI as discussed in [this comment](https://github.com/massalabs/thyra/issues/514#issuecomment-1460188271) 
    - [ ] remove tests from the Postman Collection
    - [x] remove Postman tests from the CI https://github.com/massalabs/thyra/pull/714

**QA testing**
- TBD (Please contact me if not updated so we can discuss about the best way to QA review it)
- Check that the latest CI is green

<details>
<summary>Previous version</summary>

> We wanted to use Postman as our test framework. But it clearly doesn't matches our needs as explained in [this comment](https://github.com/massalabs/thyra/issues/514#issuecomment-1468926066).

**Context**
Thyra is mostly stable and must be tested to ensure that every features are working as expected. For that, we need to add tests to Thyra's Postman collection. 

**How to**
- List the parts of Thyra that needs to be tested
- Add missing endpoints to the Postman collection
- Add tests for all endpoints in the Postman collection
- Make sure all endpoints are well tested

</details>

thomas-senechal:
[...]

PR #564, opened 2023-03-28
562 add plugin manager endpoints tests

thomas-senechal:


api/test/robot_tests/plugin_manager.robot
```diff
@@ -2,20 +2,213 @@
...
+#    ...    logo=id et sunt irure,
+#    ...    home=sunt
+#    ...    api_spec=culpa enim sint aliqua
+#    ...    url=oluptate
+#    ${response}=    POST
+#    ...    ${API_URL}/plugin-manager/register
+#    ...    expected_status=${STATUS_NO_CONTENT}
+#    ...    json=${data}
+#    ${newid}=    Get Plugin ID From Author and Name    adipisicing    ut aliqua non
+#    Should Be Equal As Strings    ${newid}    ${id}
```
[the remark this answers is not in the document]
thomas-senechal: image

For context, I especially got this error while trying to run tests 2 times in a row

(It's marked as outdated because I changed the `name`. But it's still accurate.

api/test/robot_tests/plugin_manager.robot
```diff
@@ -2,20 +2,215 @@
...
+POST /plugin-manager/{id}/execute with NotImplemented update command
+    ${id}=    Get Plugin ID From Author and Name    massalabs    hello world
+    ${data}=    Create Dictionary    command=update
+    ${response}=    POST
+    ...    ${API_URL}/plugin-manager/${id}/execute
+    ...    expected_status=${STATUS_NOT_IMPLEMENTED}
+    ...    json=${data}
+
+
+*** Keywords ***
```
thomas-senechal, line 186: @gregLibert I decided to put test specific to plugins inside this file since we shouldn't need them outside of this file.

Another way of doing it would be to create a sub directory for each part we do, and so, have the keywords specific to a part in a specific `keywords.resource` file in that subdirectory. Same for variables. We would end up with something like that:
```
- robot_tests/
   - plugins/
      - plugins.robot
      - variables.resource
      - keywords.resource
    - cmd/
      - cmd.robot
      - variables.resource
...
```

I didn't got that way because we are not sure that every parts of Thyra will need some specific variables and keywords. 

WDYT ?

api/test/robot_tests/plugin_manager.robot
```diff
@@ -2,20 +2,215 @@
...
+    ...    params=source=${source}
+    ...    expected_status=${STATUS_NO_CONTENT}
+    Sleep    1 seconds    # Wait for the plugin to be registered
+
+GET /plugin-manager with one plugin
+    ${response}=    GET    ${API_URL}/plugin-manager
+    Status Should Be    ${STATUS_OK}
+    ${listLength}=    Get Length    ${response.json()}
+    Should Be Equal As Integers    ${listLength}    1
+
```
thomas-senechal, line 33: Since the test suite is running in the written order, we can make some "scenarios" like that. We could also make them inside of a single test to make it clearer to test some specific behaviours.

api/test/robot_tests/plugin_manager.robot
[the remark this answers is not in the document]
gregLibert: You're right, this use case should be handle functionaly.

@SlnPons do you want to take it ? I can also do it if you prefer.

api/test/robot_tests/plugin_manager.robot
```diff
@@ -2,20 +2,215 @@
...
+    Basic Suite Setup
+    Delete all plugins
+
+Delete all plugins
+    Log To Console    Deleting all plugins
+    ${response}=    GET    ${API_URL}/plugin-manager
+    FOR    ${element}    IN    @{response.json()}
+        ${response}=    DELETE    ${API_URL}/plugin-manager/${element['id']}
+        Status Should Be    ${STATUS_NO_CONTENT}
+    END
```
gregLibert, line 198: This is dangerous: you will delete all the plugins. Even thoses that you didn't install. Is it on purpose?

thomas-senechal: It is on purpose. I was thinking we would need multiple plugins to test the plugin manager properly, but only one seems to be enough.

Since we only use the `hello-world` plugin for now, we could just delete it and reinstall it.

WDYT?

pkg/plugin/plugin.go
```diff
@@ -119,6 +119,7 @@ func (p *Plugin) Start() error {
 	log.Printf("Starting plugin '%s' with id %s\n", pluginName, p.ID)
 
 	p.mutex.Lock()
+	defer p.mutex.Unlock()
```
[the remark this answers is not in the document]
thomas-senechal: I made a PR for it https://github.com/massalabs/thyra/pull/570

api/test/robot_tests/plugin_manager.robot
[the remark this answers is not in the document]
SlnPons: @thomas-senechal just to be 100% sure I understand what you're talking about: 
* Alias of a plugin = name of the plugin. Is that correct? 
* we are talking about plugins that are "installed" here? (vs. plugins that are simply listed)
@gregLibert yes, please help because I have the feeling that today plugins are listed in 2 different ways (via the plugin manager and via the "plugin store repo" and both have different rules, plugins, alias, descriptions etc... 

Although, to answer simply your question: it shouldn't be possible to register / list more than one plugin with the same name. We should prevent this.

api/test/robot_tests/plugin_manager.robot
gregLibert: Ok,but should have a big warning in contributing.

api/test/robot_tests/plugin_manager.robot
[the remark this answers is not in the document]
thomas-senechal: @SlnPons
 > Alias of a plugin = name of the plugin. Is that correct?
In Thyra Plugins, an "Alias" is the combination of the Author and the Name of a plugin as follow: `{Author}/{Name}`. Those aliases are then used to redirect to the right plugin when calling `/thyra/plugin/{author}/{name}` which are used to allow users to access a plugin, and allow plugins to communicate together.

> we are talking about plugins that are "installed" here? (vs. plugins that are simply listed)
Yes, we are talking about plugins that are installed.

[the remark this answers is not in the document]
gregLibert: So you've got the rule, thyra need to reject the registry of the same plugin if it's already existing.

api/test/robot_tests/plugin_manager.robot
thomas-senechal: I already made one [in the API tests README](https://github.com/massalabs/thyra/blob/1da9a7385d2508a31d65e5b4da528fc4380b9030/api/test/README.md?plain=1#L23)

I will modify the test suite to only delete the `hello-world` plugin tho. It will be a life improvement for devs since they wont need to reinstall all their plugins after running tests.

api/test/robot_tests/plugin_manager.robot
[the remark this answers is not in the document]
thomas-senechal: To not wait for this bug to be fixed, I think we should add a task in the technical details section of  https://github.com/massalabs/thyra/issues/514 . So we are still able to merge this PR and won't forget to test this endpoint later.

WDYT @SlnPons @gregLibert ?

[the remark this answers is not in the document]
gregLibert: You should:
- comment the test linking the bug fix issue
- add in the bug fix issue to uncomment this test

api/test/robot_tests/plugin_manager/plugin_manager.robot
```diff
@@ -0,0 +1,183 @@
...
+    ...    json=${data}
+    Sleep    1 seconds    # Wait for the plugin to be restarted
+
+GET /thyra/plugin/{author}/{name}/
+    ${response}=    GET
+    ...    ${API_URL}/thyra/plugin/massalabs/hello%20world/
+    ...    expected_status=${STATUS_OK}
+    Should Contain    ${response.text}    Hello, world!
+
+# We can register multiple times the same plugin, but the aliases list isn't updated correctly.
```
gregLibert, line 73: You should add a link tho the bug fix issue here

gregLibert (approved):
:clap:

Just one non blocking suggestion.

api/test/robot_tests/plugin_manager.robot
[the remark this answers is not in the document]
thomas-senechal: There is no issue for now. Who should create it ? (I'm asking following [this comment](https://github.com/massalabs/thyra/pull/564#discussion_r1151902233) question)

api/test/robot_tests/plugin_manager/plugin_manager.robot
thomas-senechal: Done https://github.com/massalabs/thyra/pull/564/commits/71ef22d6f659639c8624069b2b75b18aa2ede750

api/test/robot_tests/cmd/websiteDeployer.wasm
Thykof, line 1: I think we can do this a better way: having a wasm into the npm js package @massalabs/sc-standards, npm init and npm i @massalabs/sc-standards 

the goal is to not git commit binary files like this one.

Thykof (approved):
LGTM, I have never work with robot framework so I can't say that much.

Thykof (changes requested):
But because this PR adds some tests, I think all github action checks must be green

api/test/robot_tests/cmd/websiteDeployer.wasm
thomas-senechal: I agree. This could be reworked in https://github.com/massalabs/thyra/issues/559 . 
WTDY @Thykof ?

ASAPSegfault (changes requested):

api/test/robot_tests/cmd/websiteDeployer.wasm
Thykof: yes, can you add a tack in that issue please?

api/test/README.md
```diff
@@ -0,0 +1,52 @@
...
+To run the tests, you need to have the following installed:
+- [Python 3](https://www.python.org/downloads/)
+- [Pip](https://pip.pypa.io/en/stable/installation/)
+- [Robot Framework](https://robotframework.org/)
+- [Requests Library](http://marketsquare.github.io/robotframework-requests/)
+
+Once you have `Python 3` and `Pip` installed, you can install the others dependencies by running the following command:
+
+```bash
+pip install -r requirements.txt
```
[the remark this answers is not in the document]
Thykof: I don't understand what you request to change

ASAPSegfault (approved):

Thykof (approved):

PR #578, opened 2023-03-30
Fix robot api test add sleep

Thykof:


.github/workflows/api.yml
```diff
@@ -54,6 +52,10 @@ jobs:
         run: xvfb-run ./main &
       - name: Define my.massa
         run: sudo echo "127.0.0.1 my.massa" | sudo tee -a /etc/hosts
+      - name: Wait for server to be up
+        uses: juliangruber/sleep-action@v1
+        with:
+          time: 25s
```
thomas-senechal, line 58: Isn't 25 seconds a bit overkill ? 😅 10 seconds should be enough, don't you think so ?

It's not blocking but we will lose a lot of time so I think we should make the sleep as short as possible and then make it longer if needed

peterjah: maybe we can ping my.massa and wait until it responds

Thykof: I tried to ping thyra on http://localhost, http://my.massa/thyra/home without success

Thykof: 15sec diff is nothing compared to uses:install/action.yml and other stuff

thomas-senechal: Let's go with this then. 
At least for now to unlock other issues.

thomas-senechal (approved):

.github/workflows/api.yml
Thykof: @peterjah see my attempt here https://github.com/massalabs/thyra/pull/490/commits/1ca36d98f5a3beab33afb592a502e8c633e36754

peterjah (approved):

PR #583, opened 2023-03-31
Rework on Thyra HomePage and plugin Manager Page

0xMazout:


0xMazout:
Considering i cannot test the linux and mac os aprt of the installation . If some one can make some tests on it por favor :)

SlnPons:
Hey fyi - from I can see from your branch, we still have some problems:
HOME PAGE: 
* elements aren't aligned 
image
* the "+ manage plugin" needs to be "manage pluginS" : spelling mistake
* the toggle is different from 1 page to another:
image
image

PLUGIN MANAGER PAGE:
* "install a plugin" bloc should be in not installed 
* the loader needs to stop loading at some point, it looks buggy otherwise.
* there is no thyra logo at the top. We need one

Maybe I don't have the latest version. If the comment above are already managed, please ignored ;) 
no need to answer, just wanted to save you some times with back and forth

0xMazout:
I don't think you are on the good branch , because all of this stuff are solved

web/home/src/pages/Home/home.tsx
```diff
@@ -67,82 +71,103 @@ function Home() {
...
+            <PluginCard
+              {...{
+                plugin: {
+                  id: plugin.id,
+                  logo: plugin.logo ? plugin.logo : massaLogomark,
+                  name: plugin ? plugin.name : 'Plugin Problem',
+                  description: plugin.description
+                    ? plugin.description
+                    : 'Plugin Problem',
+                  status: plugin.status ? plugin.status : 'Plugin Problem',
```
ASAPSegfault, line 129: Why Plugin Problem ?

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -78,10 +80,13 @@ function PluginBlock(props: PluginProps) {
     // Uninstall plugin
```
ASAPSegfault, line 80: Comment does not match function code

web/shared/styles/grid.ts
ASAPSegfault, line 1: Maybe we should extend what you did in this file with the `setResponsiveGrid` constant to other components. For example here I feel it could be useful : https://github.com/massalabs/thyra/pull/583/files#diff-cc6c260adc694cbdec38516ac595238af9243c4cab1909cdc060a6985745134dR31

ASAPSegfault (changes requested):

web/shared/styles/grid.ts
0xMazout: Yes it was a try but it's not working .. because it's not in the same React project and the render is before this read .. 

The goal is on the rebranding fusion Homepage and plugin manager page and create a UI Library to share all of this because it's very clean if we do this . I Will delete this part for this moment and take a note in the issue of the rebranding

web/home/src/pages/Home/home.tsx
0xMazout: If it's displaying this , the plugin doesn't share the infos and it's a problem , i don't know what to display in this case to show to the user there is a problem

web/home/src/index.css
```diff
@@ -23,10 +23,15 @@
...
+  width: 14px;
+  height: 14px;
+}
 
+/* Theme Toggle with icons */
 .react-toggle-track-check {
-  width: "14px";
-  height: "14px";
+  width: 14px;
+  height: 14px !important;
```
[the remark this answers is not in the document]
0xMazout: Yes i know ... it's a component not made by me , i have tried to overide this without the important and it seems there is an !important on his side,  it's the only way to overide this .. In the future we won't use this toogle

web/home/src/pages/Home/home.tsx
```diff
@@ -67,82 +71,103 @@ function Home() {
...
         console.log(err);
         // If there is an error, use the previous list
-        setPlugins(previousFetch);
+        setPlugins((previousFetch) =>
+          previousFetch.length > 0 ? previousFetch : fakePluginsList,
+        );
       });
-
+  };
+  // fetch plugins every 10 seconds
```
mazmassa, line 96: Please, let me know if you need help with that when you reach your scaling task card. 😇

web/home/src/components/pluginCard.tsx
```diff
@@ -15,21 +15,36 @@ export const PluginCard = (props: Props) => {
...
+    if (!str) {
+      return '';
+    }
+    if (str.length > length) {
+      return str.substring(0, length) + '...';
+    } else {
+      return str;
+    }
+  }
+
```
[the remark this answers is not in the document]
0xMazout: Yes truncate is a tailwind classes available , 

i had to create a class with another name : 

.minimize { overflow: hidden; text-overflow: ellipsis; }

WhiteSpace doesn't permit a return on line

web/home/src/pages/Home/home.tsx
0xMazout: Yep for sure , i think we will try to setup a longpooling on the rebranding

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -78,10 +80,13 @@ function PluginBlock(props: PluginProps) {
...
+            if (props.plugin.url === undefined) return console.error("Plugin url is undefined");
+            setisInstalling(true);
+            (await axiosServices.installPlugin(props.plugin.url)).status === (200 || 500) &&
+                setisInstalling(false);
             props.getPluginsInfo();
         } catch (error: any) {
+            setisInstalling(false);
             console.log(`Plugins failed to be downloaded , error ${error.message}`);
         }
     }
```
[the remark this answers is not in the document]
0xMazout: Maybe i didn't understand but i have tested it and it seems if my url is undefined the return is reach and my error message is printed 
and my catch is reached to when my await is failing

web/plugin-manager/src/components/pluginBlock.tsx
```diff
@@ -78,10 +80,13 @@ function PluginBlock(props: PluginProps) {
     // Uninstall plugin
     async function downloadPlugins() {
         try {
-            if(props.plugin.url === undefined) return console.error("Plugin url is undefined");
-            await axiosServices.installPlugin(props.plugin.url);
+            if (props.plugin.url === undefined) return console.error("Plugin url is undefined");
+            setisInstalling(true);
+            (await axiosServices.installPlugin(props.plugin.url)).status === (200 || 500) &&
```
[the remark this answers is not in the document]
0xMazout: Yes it's a problem you're right i have to change a way to tackle all 2xx and all 5xx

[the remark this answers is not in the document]
0xMazout: Try to rework on the answer of backend , We must have response on payload rather than with http request

mazmassa (approved):

ASAPSegfault (approved):

Issue #567, opened 2023-03-29
Windows 11 installer not working


SlnPons:
**The problem(s)**
Following the steps on the installation.md, the windows installer is not working. 

**To Reproduce**

1. Go on the last release
2. Download installer.exe
3. Open it as an admin (important note: this is not normal to ask users to do such a manip: you should just click on it!!! to be improved asap) 
4. The script run and all a sudden stops
5. Thyra is not installation
6. See logs attached for more details

**Expected behavior(s)**
User can download the installer and installer thyra without doing anything special.

**Thyra version**
v0.1.11

**Desktop (please complete the following information):**
 - OS: windows 11
 
 **Any other useful info**
[thyra_installer.log](https://github.com/massalabs/thyra/files/11099764/thyra_installer.log)

SlnPons:
cc @LucasF243 experienced this bug and can help you to reproduce if necesary.

thomas-senechal:
We should catch exceptions somewhere to avoid missing errors.

Pierregvx:
I just added windows and got it too

Pierregvx:
windows detected a virus but then it worked well when I disabled the security and launched it.

PR #584, opened 2023-04-03
567 windows 11 installer not working

Pierregvx:


thomas-senechal (approved):

Thykof:
I don't understand how the changes fix the installation

Pierregvx:
> I don't understand how the changes fix the installation

@thomas-senechal would explain it better than be but the windows decoder was crashing because of certains characters so I added a try catch to rerurn the default encoding if needed

thomas-senechal:
> > I don't understand how the changes fix the installation
> 
> @thomas-senechal would explain it better than be but the windows decoder was crashing because of certains characters so I added a try catch to reyurn the default encoding if needed

https://github.com/massalabs/thyra/issues/567#issuecomment-1488389217 Some context

scripts/installer.py
```diff
@@ -58,14 +59,17 @@ def executeCommand(self, command, shell=False, allow_failure=False) -> tuple[str
...
 
-            if allow_failure == False and process.returncode != 0:
-                self.printErrorAndExit(f"Command failed with error : {stderr.decode(locale.getpreferredencoding(False))}")
+                if allow_failure == False and process.returncode != 0:
+                    self.printErrorAndExit(f"Command failed with error : {stderr.decode(locale.getpreferredencoding(False))}")
 
-            return (stdout.decode(locale.getpreferredencoding(False)), stderr.decode(locale.getpreferredencoding(False)))
+                return (stdout.decode(locale.getpreferredencoding(False)), stderr.decode(locale.getpreferredencoding(False)))
+            except :
+                return(stdout, stderr)
```
Thykof, line 72: the bug is in the decoding of std out and str err?

Pierregvx: it was just for stderr in this case

Thykof (dismissed):

Thykof (approved):
LGTM

Thykof:
CI is red because testnet is down, this PR doesn't modify API. let's merge

PR #591, opened 2023-04-05
Debug install block and rework on grid small size screen

0xMazout:


0xMazout:
Detected a bug  after my pr was merged , on the install block this is to do a quickfix

web/home/src/pages/Home/home.tsx
```diff
@@ -136,15 +134,11 @@ function Home() {
...
       });
   };
-
-  const defineGridStyle = (length: number) => {
-    let styles = gridStyle;
-    return (styles += length <= 3 ? ' grid-cols-3 ' : setResponsiveGrid);
+  const defineGridStyle = () => {
+    const setResponsiveGrid = `max-sm:grid-cols-1 sm:grid-cols-2 md:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4`;
+    const gridStyle = 'grid grid-flow-row mx-auto mt-3 gap-4';
+    return gridStyle + ' ' + setResponsiveGrid;
```
peterjah, line 140: why a function ? it returns a constant. Just use a constant

web/plugin-manager/src/pages/Manager.tsx
```diff
@@ -105,35 +105,29 @@ function Manager() {
...
     };
 
-    const defineGridStyle = (length: number) => {
-        const numberOfCols = 3;
-        let styles = gridStyle;
-        return (styles += length <= numberOfCols ? " grid-cols-3 " : setResponsiveGrid);
+    const defineGridStyle = () => {
+        const setResponsiveGrid = `max-sm:grid-cols-1 sm:grid-cols-2 md:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4`;
+        const gridStyle = "grid grid-flow-row mx-auto mt-3 gap-4";
+        return gridStyle + " " + setResponsiveGrid;
```
peterjah, line 111: same here. no function needed

web/plugin-manager/src/components/LabelButton.tsx
```diff
@@ -20,6 +22,10 @@ const LabelButton = (props: Props) => {
     function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
         event.preventDefault();
         props.callbackToParent(value);
+        setIsPending(true);
+        setTimeout(() => {
+            setIsPending(false);
+        }, 4000);
```
peterjah, line 28: so the isPending state will just stay 4 seconds an disappear ?? :/ is it really what we want

peterjah (changes requested):

web/home/src/pages/Home/home.tsx
0xMazout: you're right it was not like this at the beginning !

web/plugin-manager/src/components/LabelButton.tsx
0xMazout: it was a dirty temporary choice but i can spend more times on this. We have to do prop drilling on 4levels for this

(it was out of the scope at the first time)

peterjah (approved):

Thykof (approved):

PR #593, opened 2023-04-06
559 add cmd endpoint tests

thomas-senechal:


thomas-senechal:
Waiting for a release of https://github.com/massalabs/thyra-plugin-wallet to fix tests. That isn't an issue for reviews.

0xMazout (approved):
Didn't tested but look good

api/test/robot_tests/cmd/keywords.resource
```diff
@@ -0,0 +1,49 @@
...
+    IF    "${pluginId}" != "${EMPTY}"
+        Log To Console    Deleting plugin with ID ${pluginId}
+        ${response}=    DELETE    ${API_URL}/plugin-manager/${pluginId}
+        Sleep    1 seconds    # Wait for the plugin to be deleted
+    END
+
+Install Wallet Plugin
+    Log To Console    Installing Massa Wallet Plugin
+    ${source}=    Set Variable
+    ...    https://github.com/massalabs/thyra-plugin-wallet/releases/download/v0.0.4/thyra-plugin-wallet_${OS}-${ARCH}.zip
```
Thykof, line 30:
```suggestion
    ...    https://github.com/massalabs/thyra-plugin-wallet/releases/download/v0.0.6/thyra-plugin-wallet_${OS}-${ARCH}.zip
```

api/test/robot_tests/cmd/websiteDeployer.wasm
Thykof, line 1: great!

api/test/testSC/assembly/contracts/main-testSC.ts
Thykof, line 1: why do we need this contract?

thomas-senechal: We need a contract to deploy. At first, I tried to simply pull the `sc-contracts` to just deploy the `websiteDeployer`, but it wasn't flexible enough. For example, we couldn't return a string with a given ID, which would allow us to check that our SC was well deployed and executed with our SC. Same for execute function. And, with this SC, we will be able to test the event-manager easily and make sure that the event we get is the right one since we will be able to send them a random ID (as I do in `executeFunction`)

And, if we need some specific behaviors for other tests, we could just add those required functions to this SC.

Thykof (approved):

Issue #511, opened 2023-03-03
Install the wallet plugin at the same time as Thyra


SlnPons:
**Context**
This issue is part of the 1st milestone to achieve for the Thyra squad working on the wallet plugin. See [here](https://www.notion.so/massa-innoteam/Wallet-85ec39b64f72481f98a0ff31c8a5608e?pvs=4#b67bc640e5414b328b64fe475658f428) for all details.

The objective of this issue is to provide the wallet plugin by default to the user.

**User flow**
1. User goes on the installation.md
2. User download the appropriate installer based on their OS 
3. User double click on the downloaded installer, and the installation starts
4. At the end of the install, user sees the icon stay in their app menu
5. User starts thyra, user can see "Massa wallet" listed in the icon tray 
6. User clicks on "Massa wallet" in the list, user sees their favorite browser opening and the wallet page opened: my.massa/thyra/massalabs/wallet

**How to**

- [ ] Update & clean if necessary thyra installation script so the wallet plugin is installed by default
- [ ] Make sure it's working on all OS (use e2e test already existing)
- [ ] Update the icon tray menu to match the user flow description above - _nice-to-have_

**Technical details**
*Give the technical insights so anyone in the team can tackle the tasks - Dev*

**QA testing**
Test it using the user flow above

peterjah:
this issue is mixing a lot of different things :/
- Install scripts on every os ,
- changes in the icon tray
- changes in api
- change the name of wallet plugin
should be splitted

SlnPons:
@peterjah you're right. 
Here is what I suggest to do: 
* change the name of the wallet plugin: will be done in another ticket ( @Thykof can you add the ticket in question please?)
* change the icon tray: nice-to-have, I'll create a ticket in the right repo to address this issue and we'll see if it relevant to do it by then. 
* install script on all OS: needed I would like the wallet to be installed at the same time as thyra. After checking, it seems that we have an endpoint that allows us to install plugin that we could use.
* change in API: not sure I understand this point... maybe a quick talk about it would be needed.

Thykof:
ticket for renaming the wallet: https://github.com/massalabs/thyra/issues/589

gregLibert:
I prefer #604: it's not Thyra job to force the installation of the plugin when it starts.

This can imply weird side effects such as:
- If I don't want to use this wallet, I can't
- If something is wrong with this wallet install, we are blocking thyra (we already had a bug on that)

SlnPons:
[...]

PR #599, opened 2023-04-07
auto install wallet

Thykof:


pkg/plugin/manager.go
```diff
@@ -246,3 +255,54 @@ func (m *Manager) Install(url string) error {
...
+			},
+			"macos-amd64": {
+				"url":
+			"https://github.com/massalabs/thyra-plugin-wallet/releases/download/v0.0.6/thyra-plugin-wallet_darwin-amd64.zip",
+				"checksum": ""
+			}
+		},
+		"version": "0.0.6",
+		"url": "https://github.com/massalabs/thyra-plugin-wallet"
+	}`
```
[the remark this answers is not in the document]
mazmassa: ... even better if we could extract this whole JSON in a specific file for that. WDYT!? 🤔

pkg/plugin/manager.go
```diff
@@ -246,3 +255,62 @@ func (m *Manager) Install(url string) error {
 
 	return nil
 }
+
+func (m *Manager) InstallWallet() error {
+	var name = "Wallet"
+	var version = "v0.0.6"
+	var baseName = "thyra-plugin-wallet"
+	var baseUrl = "https://github.com/massalabs/"
+	var url = baseUrl + baseName + "/releases/download/" + version + "/" + baseName + "_"
```
thomas-senechal, line 264: I think we should use `latest` to avoid having to update it all the time.
I don't think it's a too bad practice since, IMO, this code won't stay there. We will need something more robust to install plugins by default (we could imagine a config file listing the installed plugins, and if one plugin isn't installed, the plugin manager would install it automagically) 

```suggestion
func (m *Manager) InstallWallet() error {
	var name = "Wallet"
	var version = "latest"
	var baseName = "thyra-plugin-wallet"
	var baseUrl = "https://github.com/massalabs/"
	var url = baseUrl + baseName + "/releases/" + version + "/download/" + baseName + "_"
```

Which should give a link like:
https://github.com/massalabs/thyra-plugin-wallet/releases/latest/download/thyra-plugin-wallet_darwin-amd64.zip

pkg/plugin/manager.go
```diff
@@ -52,6 +55,12 @@ func NewManager() (*Manager, error) {
 		return manager, fmt.Errorf("while running all plugin: %w", err)
 	}
 
+	// Install plugin wallet
+	err = manager.InstallWallet()
```
thomas-senechal, line 59: Shouldn't we check if the plugin is already installed before trying to install it ?

thomas-senechal (changes requested):
Why do we have 2 differents PRs to fix this issue ? 
This one and https://github.com/massalabs/thyra/pull/604

Thykof:
> Why do we have 2 differents PRs to fix this issue ? This one and #604

because there is two way for solving this issue and the first I started #604 I don't like how I implement it and @gregLibert suggest me to try this way.

pkg/plugin/manager.go
Thykof: @SlnPons @gregLibert wdyt?

thomas-senechal: Just to make it clear, I'm not saying we need to implement the "robust" version now. 

For now, I think the link we use to download the Wallet Plugin should point to `latest` to make sure we always have the latest version of the Wallet Plugin without having to do a release of Thyra each time a new release of the Wallet Plugin gets out

Thykof: the drawback of latest is that if latest is broken, then prod is broken and we don't know it (it happened with node manager)

Thykof:
for the reasons describe here https://github.com/massalabs/thyra/issues/511#issuecomment-1505788551 I close this PR

PR #606, opened 2023-04-11
rename plugin directory

Thykof:


pkg/plugin/manager.go
```diff
@@ -20,7 +20,7 @@ import (
 // Note: the plugin directory is the /plugins inside the home directory.
 func Directory() string {
 	homeDir, _ := config.GetConfigDir()
-	pluginsDir := path.Join(homeDir, "my_plugins")
+	pluginsDir := path.Join(homeDir, "thyra-plugins")
```
thomas-senechal, line 23: IMO, it should just be `plugins` since this directory is created inside `$HOME/.config/thyra/`

```suggestion
	pluginsDir := path.Join(homeDir, "plugins")
```

Thykof: @SlnPons wdyt?

SlnPons: It looks logical as this folder will be under thyra. 
Can you confirm that it will be `thyra/plugins/...` please?

Thykof: yes it will be ~/.config/thyra/plugins/wallet-plugin

SlnPons: Ok, good for me. Thanks guys :)

thomas-senechal (approved):

peterjah (approved):

PR #641, opened 2023-05-04
623 create update mechanism to detect that an update is available

Pierregvx:


pkg/plugin/plugin.go
```diff
@@ -100,6 +109,42 @@ func (p *Plugin) SetInformation(parsedURL *url.URL) error {
...
+			return &plugin
+		}
+	}
+
+	return nil
+}
+
+func (p *Plugin) checkForPluginUpdates() (bool, error) {
+	// finf if there is an element elem in storeitems that has the same name as rootItem.Name()
+	storeItems, err := store.FetchPluginList()
```
ASAPSegfault, line 125: We already know the items comes from the store because you're calling the store module

pkg/plugin/plugin.go
```diff
@@ -100,6 +109,42 @@ func (p *Plugin) SetInformation(parsedURL *url.URL) error {
...
+	if err != nil {
+		return false, fmt.Errorf("while fetching store list: %w", err)
+	}
+
+	pluginVersion, err := version.NewVersion(p.info.Version)
+	if err != nil {
+		return false, fmt.Errorf("while parsing plugin version: %w", err)
+	}
+
+	elemInStore := findPluginByName(p.info.Name, storeItems)
```
ASAPSegfault, line 135: Why not naming it plugin instead of elemInStore ?

pkg/plugin/plugin.go
```diff
@@ -100,6 +109,42 @@ func (p *Plugin) SetInformation(parsedURL *url.URL) error {
...
+		if plugin.Name == name {
+			return &plugin
+		}
+	}
+
+	return nil
+}
+
+func (p *Plugin) checkForPluginUpdates() (bool, error) {
+	// finf if there is an element elem in storeitems that has the same name as rootItem.Name()
```
ASAPSegfault, line 124:
```suggestion
	// find if there is an element elem in storeitems that has the same name as rootItem.Name()
```

ASAPSegfault (changes requested):

pkg/plugin/plugin.go
Pierregvx: because its type is store.Plugin and not Plugin, 
indeed I could make the naming clearer

pkg/plugin/plugin.go
Pierregvx: I renamed it pluginList, WDYT ?

ASAPSegfault (approved):

pkg/plugin/plugin.go
Thykof, line 1: There is a bug here, if p.getInformation() fail, info is nil, info.URL will panic.
```go
info, err := p.getInformation()
info.URL = parsedURL
```

you can fix it with:

```go
info, err := p.getInformation()
if err != nil {
	return fmt.Errorf("error getting plugin information: %w", err)
}
info.URL = parsedURL
```

pkg/plugin/plugin.go
```diff
@@ -229,7 +274,13 @@ func New(binPath string, pluginID string) (*Plugin, error) {
...
 
 	//nolint:exhaustruct
-	plgn := &Plugin{status: Starting, BinPath: binPath + exe, ID: pluginID, quitChan: make(chan bool)}
+	plgn := &Plugin{
+		status:   Starting,
+		BinPath:  binPath + exe,
+		ID:       pluginID,
+		quitChan: make(chan bool),
+	}
+	plgn.info, _ = plgn.getInformation()
```
Thykof, line 283: you should not ignore the error

```suggestion
	plgn.info, err = plgn.getInformation()
	if err != nil {
		return nil, fmt.Errorf("getting plugin information: %w", err)
	}
```

pkg/plugin/plugin.go
```diff
@@ -229,7 +274,13 @@ func New(binPath string, pluginID string) (*Plugin, error) {
 	}
 
 	//nolint:exhaustruct
-	plgn := &Plugin{status: Starting, BinPath: binPath + exe, ID: pluginID, quitChan: make(chan bool)}
+	plgn := &Plugin{
+		status:   Starting,
+		BinPath:  binPath + exe,
```
Thykof, line 279: no blocking, I would rename `exe` var to `ext`

Thykof (changes requested):

pkg/plugin/plugin.go
```diff
@@ -100,6 +109,42 @@ func (p *Plugin) SetInformation(parsedURL *url.URL) error {
 	return nil
 }
 
+func findPluginByName(name string, plugins []store.Plugin) *store.Plugin {
+	// for each plugin in the plugins list, check if the name matches the name of the plugin
+	for _, plugin := range plugins {
+		if plugin.Name == name {
```
peterjah, line 115: can you modify this, to check the name AND the author ??
It will allow us more flexibilité (by instance having 2 plugins with same name but differents author)

peterjah (changes requested):

pkg/plugin/plugin.go
```diff
@@ -79,13 +82,24 @@ func (p *Plugin) getInformation() (*Information, error) {
...
 	info, err := p.getInformation()
+	if err != nil {
+		return fmt.Errorf("error getting plugin information: %w", err)
+	}
+
 	info.URL = parsedURL
 
 	if err != nil {
 		return fmt.Errorf("error getting plugin information: %w", err)
 	}
```
Thykof, line 95: now you check err 2 times ^^

```suggestion
```

pkg/plugin/plugin.go
```diff
@@ -100,6 +114,42 @@ func (p *Plugin) SetInformation(parsedURL *url.URL) error {
...
+		return false, fmt.Errorf("while fetching store list: %w", err)
+	}
+
+	pluginVersion, err := version.NewVersion(p.info.Version)
+	if err != nil {
+		return false, fmt.Errorf("while parsing plugin version: %w", err)
+	}
+
+	pluginInStore := findPluginByName(p.info.Name, pluginList)
+	if pluginInStore != nil { // if there is a plugin with the same name, check if the version is greater than the current one
```
Thykof:
```suggestion
	// If there is a plugin with the same name,
	// check if the version is greater than the current one.
	if pluginInStore != nil {
```

Thykof (changes requested):

PR #644, opened 2023-05-04
507 version always returns main

Pierregvx:


peterjah (approved):

.github/workflows/api.yml
```diff
@@ -16,8 +16,10 @@ jobs:
         uses: ./.github/actions/install
         with:
           os: ubuntu-latest
+      - name: fetch tags
+        run: git fetch --tags
       - name: Update version
-        run: echo -e "${GITHUB_REF_NAME}" > cmd/thyra-server/version.txt
+        run: git describe --tags --abbrev=0 --always > cmd/thyra-server/version.txt
```
peterjah, line 22: i propose to add a suffix to the version with the short hash of the current commit
this will allow us to differentiate a release version from a dev version

peterjah (changes requested):

.github/workflows/build.yml
```diff
@@ -45,8 +45,9 @@ jobs:
         uses: ./.github/actions/install
         with:
           os: ${{ matrix.os }}
-      - name: Update version
-        run: echo -e "${GITHUB_REF_NAME}" > cmd/thyra-server/version.txt
+      - name: Update version for a release
+        if: ${{ inputs.tag_name != '' }}
+        run: echo -e "${{ inputs.tag_name }}" > cmd/thyra-server/version.txt
```
peterjah, line 50: you have to handle the else case here. if not the version will be empty for dev builds

Pierregvx: I did that at the beginning but there is only a build for the installer (and not the server) in the workflow and as you mentioned me this  morning, the installer uses the server from the latest release [this is the link used](https://github.com/massalabs/thyra/blob/a5642d5efb6a31c18745324c3f50b57fc46c661b/scripts/linux_installer.py#L19-L20) and not the one built from the branch
therefore if I am not wrong updating the version.txt for a build won't do anything with the current workflow, WDYT?

peterjah: yes my remark is about the dev builds, not the release which is included in installer.
You can ignore this, we will see later if some improvments are needed

peterjah (approved):

Thykof (approved):

PR #645, opened 2023-05-04
refactor pkg: make it reusable

Thykof:


peterjah (approved):
Good job!
Instead of prepareAddress, i think it would be more accurate to use serializeAddress.

SlnPons:
Can't test, no details.

Thykof:
> Can't test, no details.

it only refactoring the code, I will test no-reg by deploying a flappy on buildnet

Thykof:
here we go! https://flappy.massa/ on buildnet

PR #648, opened 2023-05-09
added update endpoint

Pierregvx:
added the endpoints for the plugin update, 
the update function in implemented here #627


Thykof (changes requested):

peterjah (changes requested):
There is no need for a new endpoint.
just use `POST /plugin-manager/%s/execute` with the option "update"

peterjah (approved):

Thykof (approved):

Thykof (changes requested):
You need to change "update" to something else because now it's implemented https://github.com/massalabs/thyra/blob/main/api/test/robot_tests/plugin_manager/plugin_manager.robot#L181 ^^

api/test/robot_tests/plugin_manager/plugin_manager.robot
```diff
@@ -178,7 +178,7 @@ POST /plugin-manager/register with invalid body
 
 POST /plugin-manager/{id}/execute with NotImplemented update command
```
peterjah, line 179: with NotImplemented update command => with unknown command

Thykof:
> You need to change "update" to something else because now it's implemented https://github.com/massalabs/thyra/blob/main/api/test/robot_tests/plugin_manager/plugin_manager.robot#L181 ^^

my bad, you can remove this test, test for invalid command already exists

Thykof (approved):

PR #649, opened 2023-05-09
627 implement the update mechanism

Pierregvx:


pkg/plugin/manager.go
```diff
@@ -225,29 +225,84 @@ func (m *Manager) Install(url string) error {
...
+
+	return nil
+}
+
+func (m *Manager) Update(correlationID string) error {
+	pluginsDir := Directory()
+
+	plgn, err := m.Plugin(correlationID)
+	if err != nil {
+		return fmt.Errorf("deleting plugin %s: %w", correlationID, err)
```
Thykof, line 274: you are not deleting the plugin here, you are getting the plugin from the node manager.

pkg/plugin/manager.go
```diff
@@ -225,29 +225,84 @@ func (m *Manager) Install(url string) error {
...
+	}
+
+	pluginInStore := findPluginByName(plgn.info.Name, pluginList)
+
+	url, _, _, err := pluginInStore.GetDLChecksumAndOs()
+	if err != nil {
+		return fmt.Errorf("while fetching plugin %s: %w", plgn.info.Name, err)
+	}
+
+	err = os.RemoveAll(filepath.Join(pluginsDir, plgn.info.Name))
```
Thykof, line 293: why don't you use Delete function?
why don't you stop the plugin before removing the binary?

pkg/plugin/manager.go
```diff
@@ -225,29 +225,84 @@ func (m *Manager) Install(url string) error {
...
+	}
+
+	return pluginPath, nil
+}
+
+// Install grabs a remote plugin from the given url and install it locally.
+func (m *Manager) Install(url string) error {
+	pluginPath, err := m.DownloadPlugin(url, true)
+	if err != nil {
+		return fmt.Errorf("installing plugin at %s: %w", url, err)
```
Thykof, line 258:
```suggestion
		return fmt.Errorf("downloading plugin at %s: %w", url, err)
```

pkg/plugin/manager.go
```diff
@@ -225,29 +225,84 @@ func (m *Manager) Install(url string) error {
...
+	if err != nil {
+		return fmt.Errorf("while fetching plugin %s: %w", plgn.info.Name, err)
+	}
+
+	err = os.RemoveAll(filepath.Join(pluginsDir, plgn.info.Name))
+	if err != nil {
+		return fmt.Errorf("updating plugin %s: %w", plgn.info.Name, err)
+	}
+
+	_, err = m.DownloadPlugin(url, false)
```
Thykof, line 298: why don't you call Install function?

Thykof (changes requested):
I think you should re start the plugin, stop and run.

pkg/plugin/manager.go
Thykof: because we don't want to initialize a new plugin

pkg/plugin/manager.go
Thykof: to not delete wallet accounts for example

Thykof:
to fix lint errors you can have a look at https://github.com/massalabs/thyra/blob/main/CONTRIBUTING.md#how-to-resolve-golangci-lint-recurring-errors-

peterjah (approved):

Pierregvx:
there is an issue with the github actions rn https://www.githubstatus.com/

Thykof (approved):

PR #654, opened 2023-05-11
 650-check-if-the-store-version-is-compliant-with-the-thyra-one-in-the-update-mechanism

Pierregvx:


Thykof (approved):

peterjah:
It would be nice to have unit tests on this. and test agains differents version syntaxes (0.0.1, 1.2.3-dev, 1.3 .. etc

Thykof:
@peterjah we will add unit test in another PR, can we merge this one it's ok for you?

Thykof (approved):

pkg/store/store.go
```diff
@@ -42,18 +47,20 @@ const (
 	cacheExpirationMinutes = 30
```
peterjah, line 47: the cache has been removed. delete all mention of it

pkg/plugin/manager.go
```diff
@@ -45,9 +45,9 @@ type Manager struct {
 }
 
 // NewManager instantiates a manager struct.
-func NewManager(storeMS *store.Store) (*Manager, error) {
-	//nolint:exhaust,exhaustruct
-	manager := &Manager{plugins: make(map[string]*Plugin), authorNameToID: make(map[string]string), store: storeMS}
+func NewManager() (*Manager, error) {
+	//nolint:exhaust,exhaustruct,lll
+	manager := &Manager{plugins: make(map[string]*Plugin), authorNameToID: make(map[string]string), store: store.StoreInstance}
```
peterjah, line 50: i don't think you need to give the store instance to the manager here. you can just call it from everywhere its needed

pkg/store/store.go
```diff
@@ -42,18 +47,20 @@ const (
...
 )
 
-func NewStore() (*Store, error) {
+func NewStore() error {
 	//nolint:exhaustruct
 	storeMassaStation := &Store{}
```
peterjah, line 52: you can assign StoreInstance directly here. if there is an error, the app will crash anyway

peterjah (changes requested):

peterjah (approved):

PR #657, opened 2023-05-12
655 move config directory

thomas-senechal:
The installer part isn't really clean, but since this installer will be removed soon, it should be fine.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -3,19 +3,16 @@ package config
 import (
 	"errors"
 	"os"
-	"path"
 )
 
 //nolint:gochecknoglobals
 var Version = "dev"
 
 func GetConfigDir() (string, error) {
-	homeDir, err := os.UserHomeDir()
+	confDir, err := getConfigDir()
 	if err != nil {
-		return "", errors.New("Unable to get user home dir: " + err.Error())
+		return "", err
 	}
-
-	confDir := path.Join(homeDir, ".config", "thyra")
 
 	_, err = os.Stat(confDir)
 	if err != nil {
@@ -23,4 +20,18 @@ func GetConfigDir() (string, error) {
 	}
 
 	return confDir, nil
+}
+
+func GetCertDir() (string, error) {
+	certDir, err := getCertDir()
+	if err != nil {
+		return "", err
+	}
+
+	_, err = os.Stat(certDir)
+	if err != nil {
+		return "", errors.New("Unable to read cert dir: " + certDir + ": " + err.Error())
+	}
+
+	return certDir, nil
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -8,6 +8,7 @@ import (
 //nolint:gochecknoglobals
 var Version = "dev"
 
+// GetConfigDir returns the config directory for the current OS.
 func GetConfigDir() (string, error) {
 	confDir, err := getConfigDir()
 	if err != nil {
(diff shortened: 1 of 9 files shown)


peterjah (approved):

pkg/config/config.go
```diff
@@ -3,24 +3,37 @@ package config
...
+// GetCertDir returns the cert directory for the current OS.
+func GetCertDir() (string, error) {
+	certDir, err := getCertDir()
+	if err != nil {
+		return "", err
+	}
+
+	_, err = os.Stat(certDir)
+	if err != nil {
+		return "", errors.New("Unable to read cert dir: " + certDir + ": " + err.Error())
```
Thykof, line 35:
```suggestion
		return "", fmt.Errorf("Unable to read cert directory: %s: %w", certDir, err)
```

pkg/config/config.go
```diff
@@ -3,24 +3,37 @@ package config
...
 	if err != nil {
-		return "", errors.New("Unable to get user home dir: " + err.Error())
+		return "", err
 	}
 
-	confDir := path.Join(homeDir, ".config", "thyra")
-
 	_, err = os.Stat(confDir)
 	if err != nil {
 		return "", errors.New("Unable to read config dir: " + confDir + ": " + err.Error())
```
Thykof, line 20:
```suggestion
		return "", fmt.Errorf("Unable to read config directory: %s: %w", confDir, err)
```

pkg/config/config_unix.go
```diff
@@ -0,0 +1,38 @@
...
+	return path, nil
+}
+
+// getCertDir returns the cert directory for the current OS.
+// On Unix, it is placed in /etc/massa/certs.
+func getCertDir() (string, error) {
+	path := path.Join("/", "etc", "massastation", "certs")
+
+	return path, nil
+}
```
Thykof, line 38: why don't you try to create the dir if not exists already?

pkg/config/config_windows.go
```diff
@@ -0,0 +1,34 @@
...
+
+	return dir, nil
+}
+
+// getCertDir returns the cert directory for the current OS.
+// On Windows, it is at the same location as the executable, by default `C:\Program Files(x86)\MassaStation\certs`.
+func getCertDir() (string, error) {
+	confDir, err := getConfigDir()
+	if err != nil {
+		return "", err
```
Thykof, line 28: better to be consistent in the error messages.

```suggestion
		return "", fmt.Errorf("getting executable path: %w", err)
```

scripts/installer.py
Thykof, line 1: in this file you execute command like sudo chmod... but how are you sure it will not be executed on a windows OS?

I am not sure that: `self.SUDO_INSTALLATION = False` is clear and maintainable enough.

In the generic installer script I would use only generic commands python instructions.

Thykof (changes requested):

pkg/config/config_unix.go
thomas-senechal: Creating it requires `sudo`. So we cannot create it if it doesn't exists.

scripts/installer.py
thomas-senechal: As explained in the description of this PR, the installation scripts will be deleted soon, so I think it's okay to go with that `SUDO_INSTALLATION` variable even if it's not very clear nor maintainable.

pkg/config/config_windows.go
thomas-senechal: We are not trying to get the executable path in that case. But yes, we need to add some information about the fact that we failed to get the certs folder path.

PR #662, opened 2023-05-15
661 move the new windows installer

thomas-senechal:


installer/windows/build_installer.py
```diff
@@ -0,0 +1,348 @@
...
+    )
+    args = parser.parse_args()
+
+    VERSION = args.version
+    FORCE_BUILD = args.force_build
+
+    if not sys.platform.startswith("win"):
+        if not FORCE_BUILD:
+            print(
+                "This script is only meant to be run on Windows. If you wish to run it anyway, use the '-f' flag."
```
Thykof, line 342: just asking, in what case would we like to force build?

installer/windows/build_installer.py
```diff
@@ -0,0 +1,348 @@
...
+        default=FORCE_BUILD,
+        help="Force the build on non-Windows platforms.",
+    )
+    args = parser.parse_args()
+
+    VERSION = args.version
+    FORCE_BUILD = args.force_build
+
+    if not sys.platform.startswith("win"):
+        if not FORCE_BUILD:
```
Thykof, line 340:
```suggestion
        if not args.force_build:
```

installer/windows/build_installer.py
```diff
@@ -0,0 +1,348 @@
...
+        "-f",
+        "--force",
+        action="store_true",
+        dest="force_build",
+        default=FORCE_BUILD,
+        help="Force the build on non-Windows platforms.",
+    )
+    args = parser.parse_args()
+
+    VERSION = args.version
```
Thykof, line 336: it might not be a good practice to modify global variable, I would have pass the value to the functions build_installer and create_wxs_file

installer/windows/build_installer.py
```diff
@@ -0,0 +1,348 @@
...
+    FORCE_BUILD = args.force_build
```
Thykof, line 337:
```suggestion
```

installer/windows/build_installer.py
```diff
@@ -0,0 +1,348 @@
...
+import argparse
+import os
+import shutil
+import subprocess
+import sys
+import urllib.request
+import zipfile
+
+BUILD_DIR = "buildmsi"
+FORCE_BUILD = False
```
Thykof, line 16: see comment bellow

```suggestion
```

installer/windows/build_installer.py
Thykof, line 1: there is not try catch around main function, what happen if install_dependencies fail? do we want to handle this kind of error? If no, ignore my comment

Thykof (changes requested):
here is my review of python code

peterjah (approved):

Thykof (approved):

PR #663, opened 2023-05-15
Switch test to buildnet

Thykof:


peterjah (commented):
So Innonet is over ?? i think youcan remove every occurence of it
pkg/config/network.go
cmd/thyra-server/main.go
api/test/massa_endpoints.postman_collection.json

Thykof:
> So Innonet is over ?? i think youcan remove every occurence of it pkg/config/network.go cmd/thyra-server/main.go api/test/massa_endpoints.postman_collection.json

You are right, I have create issue #664 , let's merge this one, it's getting urgent.

PR #681, opened 2023-05-26
675 move ci windows installer

thomas-senechal:


.github/workflows/build.yml
```diff
@@ -95,7 +98,39 @@ jobs:
...
           retention-days: 7
+
+  build-packages-windows:
+    name: Build .msi package for amd64
+    needs: build
+    defaults:
+      run:
+        working-directory: ./installer
+
+    runs-on: windows-latest
```
Thykof, line 113: `runs-on: windows-latest` it's not a good practice to rely on latest versions (same for macos). Is there a specific reason?

installer/windows/build_installer.py
```diff
@@ -64,6 +81,10 @@ def build_massastation():
         ],
         check=True,
     )
+    os.rename(
+        os.path.join("..", "cmd", "massastation", "MassaStation.exe"),
```
Thykof, line 85: here why it's `MassaStation.exe` and not var `MASSASTATION_BINARY`?

Thykof (changes requested):

installer/windows/build_installer.py
thomas-senechal: When we run `fyne package`, the `-name` correspond to the name of the package we want to make. In the case of Windows, the resulting package is a `.exe` named as the `-name` option, so in that case, `MassaStation.exe`. That's why we need to move it, and rename it.

.github/workflows/build.yml
thomas-senechal: There is no real reason for that.
Do you think I should use a specific version ? Or is the latest windows version fine (I'm talking about windows-2022) ?

installer/windows/build_installer.py
Thykof: okay, you can add a comment in the code explaining that.

.github/workflows/build.yml
Thykof: @peterjah WDYT?

installer/windows/build_installer.py
thomas-senechal: Done

Thykof (approved):
LGTM

.github/workflows/build.yml
gregLibert: No, we should use lastest.

.github/workflows/build.yml
```diff
@@ -95,7 +98,39 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
```
gregLibert, line 115: Set a name to this step.

gregLibert (approved):

Thykof (approved):

PR #683, opened 2023-05-26
675 move ci macos installer

thomas-senechal:


.github/workflows/build.yml
```diff
@@ -42,7 +44,10 @@ jobs:
           export CGO_ENABLED=1
           export CC=${{ matrix.cc }}
           go build -ldflags "-X github.com/massalabs/thyra/pkg/config.Version=${{env.VERSION}}" -o massastation_${{ matrix.target }}_${{ matrix.arch }} ./cmd/massastation/
-      - name: Upload Package
+          if [ ${{ matrix.target }} != "windows" ]; then
+            fyne package -name MassaStation -icon ./cmd/massastation/logo.png -appID net.massalabs.massastation -exe massastation_${{ matrix.target }}_${{ matrix.arch }}
+          fi
```
peterjah, line 49: maybe you can put this in a different step of the workflow ? with a proper name

thomas-senechal: The issue is that, if we move this to another workflow, we need to reinstall all the dependencies, which takes like 5min per OS, and we end up with a CI that needs 15min to run (we are already at 10min for the `api` workflow that does the build and packaging).
We could reduce a bit the time the CI takes to pass by removing the installer building, but we are planning to try to test the installers in the CI.

peterjah: i just said differnt step, but same workflow

thomas-senechal: Oh sorry! 
I kept those in the same step to avoid having to redo the exports.
I will check if it's possible to do it in a clean way

peterjah (approved):

.github/workflows/build.yml
```diff
@@ -34,15 +34,24 @@ jobs:
...
       - name: Build binary for ${{ matrix.target }} on ${{ matrix.arch}}
         shell: bash
         run: |
-          export GOOS=${{ matrix.target }}
-          export GOARCH=${{ matrix.arch }}
-          export CGO_ENABLED=1
-          export CC=${{ matrix.cc }}
           go build -ldflags "-X github.com/massalabs/thyra/pkg/config.Version=${{env.VERSION}}" -o massastation_${{ matrix.target }}_${{ matrix.arch }} ./cmd/massastation/
-      - name: Upload Package
+      - name: Package binary for ${{ matrix.target }} on ${{ matrix.arch}}
```
Thykof, line 49: there is no 'Package binary for' for windows

.github/workflows/build.yml
```diff
@@ -51,3 +60,42 @@ jobs:
...
           retention-days: 7
+      - name: Upload Package
+        uses: actions/upload-artifact@v3
+        with:
+          name: massastation_${{ matrix.target }}_${{ matrix.arch }}_package
+          path: |
+            MassaStation.app
+            MassaStation.tar.xz
+          if-no-files-found: warn
+          retention-days: 1
```
Thykof, line 71: why don't we use same value as others?

```suggestion
          if-no-files-found: error
          retention-days: 7
```

Thykof (changes requested):

.github/workflows/build.yml
thomas-senechal: The `warn` is there only while I'm working on this part of the CI. It will be replaced by an error in next PRs.
And the retention-days is shorter because we need those artifacts only to build the installers. It's basically to save storage as much as possible

Thykof (approved):

PR #705, opened 2023-05-30
675 fix impossible to open error

thomas-senechal:


installer/macos/create_pkg.sh
```diff
@@ -33,21 +34,28 @@ install_massastation_build_dependencies() {
 build_massastation() {
     install_massastation_build_dependencies
 
-    go generate ../... || fatal "go generate failed for $MASSASTATION_BINARY_NAME"
+    go generate ../... || fatal "go generate failed for $MASSASTATION_INSTALLER_NAME"
```
peterjah, line 37: are you sure about this? `go generate ../...`, for generating all directives using global path its `go generate ./...`

thomas-senechal: Yes. Because this script is supposed to be started from the `installer` directory

peterjah: just tried! it looks its working well :)

PR #706, opened 2023-05-31
675 fix impossible to open error 2

thomas-senechal:


peterjah (changes requested):
for windows you can set the .exe extension in the matrix like its done here:
 https://github.com/massalabs/thyra-plugin-wallet/blob/625f017947aa14c306a3206b02ad0b383724474c/.github/workflows/build.yml#LL30C14-L30C22
 it will allow you to handle all cases in one

peterjah (approved):

.github/workflows/build.yml
```diff
@@ -65,13 +64,14 @@ jobs:
           if-no-files-found: error
           retention-days: 7
       - name: Upload Package
+        if: matrix.target != 'windows'
```
Thykof, line 67: why don't you upload package for windows? Add a comment for the specific reason please.

Thykof (changes requested):

Thykof (approved):

PR #718, opened 2023-06-05
715 update the websites page to add description

B-Naoufal:


pkg/config/network.go
```diff
@@ -15,7 +15,7 @@ const (
 	labnetDNSAddress = "AS1PV17jWkbUs7mfXsn8Xfs9AK6tHiJoxuGu7RySFMV8GYdMeUSh"
 
 	buildnetNodeURL    = "https://buildernet.massa.net/api/v2"
-	buildnetDNSAddress = "AS1HqqZF5nFiZEzC7A19t7pUBRtvNfAq5c5PsESSPRE9eFYrGxhg"
+	buildnetDNSAddress = "AS12j2rnuorLQXq9ysbJSpHPKtDr94A2J7Xby78aEGSH3m12cZkQH"
```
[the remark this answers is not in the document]
B-Naoufal: Yes i deployed a new dns contract which supports the description

api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -861,6 +861,13 @@ paths:
           required: true
           x-nullable: false
           description: URL without dot (.), upper case and special characters
```
peterjah, line 851: can you update this description with the corresponding allowed names format?

peterjah: its [A-Za-z0-9]

int/api/websites/all.go
```diff
@@ -64,23 +64,28 @@ func Registry(config config.AppConfig) ([]*models.Registry, error) {
...
 			return nil, fmt.Errorf("reading key '%s' at '%s': %w", metaKey, websiteStorerAddress, err)
 		}
 
 		registry[index] = &models.Registry{
-			Name:     convert.BytesToString(websiteNames[index]), // name of website : flappy.
-			Address:  websiteStorerAddress,                       // website Address
-			Metadata: websiteMetadata.CandidateValue,             // website metadata.
+			Name:        convert.BytesToString(websiteNames[index]), // name of website : flappy.
+			Address:     websiteStorerAddress,                       // website Address
+			Description: websiteDescription,                         // website Description
```
peterjah, line 87: this code is very ugly :/ can you refactor a bit?
idea from chat GPT :)
```
	for index, dnsValue := range dnsValues {
		websiteValue := convert.ByteToStringArray(dnsValue.CandidateValue)

		websiteStorerAddress := websiteValue[0]
		websiteDescription := websiteValue[2]

		websiteMetadata, err := node.DatastoreEntry(client, websiteStorerAddress, convert.StringToBytes(metaKey))
		if err != nil {
			return nil, fmt.Errorf("reading key '%s' at '%s': %w", metaKey, websiteStorerAddress, err)
		}

		registry[index] = &models.Registry{
			Name:        convert.BytesToString(websiteNames[index]),
			Address:     websiteStorerAddress,
			Description: websiteDescription,
			Metadata:    websiteMetadata.CandidateValue,
		}
	}
```

pkg/my/domain.go
```diff
@@ -73,8 +73,12 @@ func Websites(config config.AppConfig, client *node.Client, domainNames []string
 
 	for i := 0; i < len(domainNames); i++ { //nolint:varnamelen
 		contractAddressesIndex := 0
+		websiteDescriptionIndex := 2
+
 		contractAddress := convert.ByteToStringArray(contractAddresses[i].CandidateValue)[contractAddressesIndex]
 
+		websiteDescription := convert.ByteToStringArray(contractAddresses[i].CandidateValue)[websiteDescriptionIndex]
+
```
peterjah: similar refacto can be done here

peterjah (changes requested):

api/swagger/server/restapi/resource/swagger.yml
thomas-senechal: Shouldn't it be `[a-z0-9]` ? 
The description says "without upper case"

thomas-senechal: There is a kind of validation on MassaStation side : https://github.com/massalabs/thyra/pull/718/files#diff-3e09e4ebd760f825d1cf5194f769c18c2b1d1c5f88e47e9ab74bea2498de3ebdL848 
I guess Swagger will block any request with `name` that doesn't match that Regex

And, I don't think we want to allow users to use uppercase in their website names. Otherwise, we will endup with:
- `Abc.massa`
- `abc.massa`
- `aBc.massa`
- ...

int/api/html/front/website.html
```diff
@@ -180,6 +188,7 @@ <h3 class="stepper-title"></h3>
                         <thead>
                             <tr>
                                 <th scope="col">Website name</th>
+                                <th scope="col">Website Description</th>
```
thomas-senechal, line 191: Why "Description" and not "description" ? 
Either we use uppercase everywhere, or lowercase everywhere:
```suggestion
                                <th scope="col">Website Name</th>
                                <th scope="col">Website Description</th>
```

```suggestion
                                <th scope="col">Website name</th>
                                <th scope="col">Website description</th>
```

thomas-senechal (approved):

PR #721, opened 2023-06-06
Fix brew install check bug

thomas-senechal:
- Fixes a bug with the `if` for homebrew check -> The previous way couldn't work anymore since we now have `set -e` that exits at the first error.
- Fixes bugs related to brew installation


Thykof (approved):
LGTM

gregLibert (approved):

PR #731, opened 2023-06-08
add massastation frontend skeleton

Thykof:


SlnPons:
Niceeeeee @Thykof ! A real FE dev now ;)

PR #749, opened 2023-06-13
use cache for apt in workflows

peterjah:
QA: check that CI is green


Thykof:
`Unable to reserve cache with key Linux-apt-build-v1, another job may be creating this cache.`

Thykof:
it exists several reusable action that handle the cache of dependencies.

thomas-senechal:
I'm closing this issue as it's staled since months and we don't really need it anymore

PR #752, opened 2023-06-13
Set up mirage server

Pierregvx:


pivilartisant (approved):

web/massastation/src/mirage/.gitkeep
maxime-carabina, line 1: You can remove this file because mirage folder is not empty now.

maxime-carabina (approved):

mazmassa (approved):

PR #753, opened 2023-06-14
update plugin-store & manager responses

Pierregvx:


pivilartisant (approved):

api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -1050,12 +1055,18 @@ definitions:
...
+        type: string
       description:
         type: string
       version:
         type: string
       url:
         type: string
+      logo:
+        type: string
+      massastationVersion:
```
peterjah, line 1068: its weird to find massaStationVersion here.
it is not a property of a plugin.
Or maybe its the massaStation minimum version required ? in that case its poorly named

peterjah (changes requested):
to follow

api/swagger/server/restapi/resource/swagger.yml
Pierregvx: it is part of th manifest https://github.com/massalabs/thyra-plugin-store/blob/04a13e300f78a704e23308413ed8a6315ce68e4e/plugins.json#L5
but not really used in the front end, I can remove it

peterjah: ok! so its the minimum required version. can you just rename this ?
by instance `massastationMinVersion`

peterjah (approved):

PR #781, opened 2023-06-18
Implement XSS protection In Description

B-Naoufal:
FYI : https://developers.redhat.com/articles/2022/06/28/cross-site-scripting-explanation-and-prevention-go#vulnerable_code_examples

### DIFF
--- a/int/api/websites/all.go
+++ b/int/api/websites/all.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"text/template"
 
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/massalabs/thyra/api/swagger/server/models"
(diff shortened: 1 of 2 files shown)


int/api/websites/all.go
```diff
@@ -78,6 +79,9 @@ func Registry(config config.AppConfig) ([]*models.Registry, error) {
 		websiteStorerAddress := websiteValue[indexOfWebsiteAddress]
 		websiteDescription := websiteValue[indexOfWebsiteDescription]
 
+		// Prevent XSS by escaping special characters in websiteDescription
+		escapedDescription := template.HTMLEscapeString(websiteDescription)
+
```
peterjah, line 84: can you factorize this code  in a dedicated function? and use it in the second place its needed

peterjah (approved):

int/api/websites/all.go
B-Naoufal: https://github.com/massalabs/station/issues/792

B-Naoufal:
covered by https://github.com/massalabs/station/pull/793

PR #788, opened 2023-06-19
fix mirage

pivilartisant:
We fix the bug created by the base url variable that breaks mirage

### DIFF
--- a/web/massastation/src/main.tsx
+++ b/web/massastation/src/main.tsx
@@ -17,9 +17,10 @@ import { Error, NotFound, Index, Base, Store } from './pages/index';
 import mockServer from './mirage/server.js';
 
 const baseURL = import.meta.env.VITE_BASE_APP;
+const baseENV = import.meta.env.VITE_ENV;
 
-if ([ENV.DEV, ENV.TEST].includes(baseURL)) {
-  mockServer(baseURL);
+if ([ENV.DEV, ENV.TEST].includes(baseENV)) {
+  mockServer(baseENV);
 }
 
 const queryClient = new QueryClient();


web/massastation/src/main.tsx
```diff
@@ -18,8 +18,9 @@ import mockServer from './mirage/server.js';
 
 const baseURL = import.meta.env.VITE_BASE_APP;
 
-if ([ENV.DEV, ENV.TEST].includes(baseURL)) {
-  mockServer(baseURL);
+// Add ENV.STANDALONE to the array to enable MirageJS
+if ([ENV.DEV, ENV.TEST].includes(import.meta.env.VITE_ENV)) {
+  mockServer(import.meta.env.VITE_ENV);
```
mazmassa, line 23: Can we have something similar to the `baseURL` to use once down? baseENV or related...

```suggestion
if ([ENV.DEV, ENV.TEST].includes(baseENV)) {
  mockServer(baseENV);
```

Pierregvx (approved):

web/massastation/src/main.tsx
```diff
@@ -17,9 +17,11 @@ import { Error, NotFound, Index, Base, Store } from './pages/index';
 import mockServer from './mirage/server.js';
 
 const baseURL = import.meta.env.VITE_BASE_APP;
+const baseENV = import.meta.env.VITE_ENV;
 
-if ([ENV.DEV, ENV.TEST].includes(baseURL)) {
-  mockServer(baseURL);
+// Add ENV.STANDALONE to the array to enable MirageJS
```
Thykof, line 22: there is no standalone mode for MassaStation, this comment was specific to the wallet plugin


```suggestion
```

pivilartisant: @maxime-carabina can answer :)

Thykof: VITE_ENV: test or dev or prod
VITE_BASE_APP: base path URL of the application

mazmassa: Thanks @Thykof

PR #804, opened 2023-06-19
fix web app routing

Thykof:


web/massastation/src/main.tsx
```diff
@@ -27,8 +28,13 @@ const queryClient = new QueryClient();
 const router = createBrowserRouter(
   createRoutesFromElements(
     <Route path={baseURL} element={<Base />}>
+      <Route path="" element={<Index />} />
+
+      {/* go swagger will redirect to index.html to serve index.html file */}
+      {/* here we match index.html en navigate to "" which is Index */}
```
thomas-senechal, line 34: I guess you meant
```suggestion
      {/* here we match "index.html" to navigate to "" which is Index */}
```
`en` -> `to`

peterjah (approved):

thomas-senechal (approved):

PR #813, opened 2023-06-20
add parameters the usepost mutation

Pierregvx:


web/massastation/src/custom/api/usePost.ts
```diff
@@ -8,15 +8,28 @@ import { UseMutationResult, useMutation } from '@tanstack/react-query';
 
 export function usePost<T, P = unknown>(
   resource: string,
-): UseMutationResult<P, unknown, T, unknown> {
-  var url = `${import.meta.env.VITE_BASE_API}/${resource}`;
+): UseMutationResult<
+  P,
+  unknown,
+  { params?: Record<string, string>; payload?: T },
```
Thykof, line 14: can you explain this?

Thykof (changes requested):
why don't you simply concatenate the query params in the ressource argument before calling the mutate?

web/massastation/src/custom/api/usePost.ts
Pierregvx: these are the arguments that you give when we call the mutation,
ex
```ts
    const payload = { name: 'John Doe', age: 25 };
    const params = { token: 'abc123' };

    mutate({  payload, params });
```

mazmassa (approved):
We need to revisit that to not have a custom usePost for massastation. But we'll do in future.

mazmassa:
> why don't you simply concatenate the query params in the ressource argument before calling the mutate?

Yep! I thought the same. But he needs a refreshed mutation. I mean, a mutation already with the correct url is filled on the field.

For now we have a custom format of usePost in massastation. (Weird!!!). But we have no time to do it differently.

PR #816, opened 2023-06-20
Upload website

Thykof:
known issues:
- there is no spinner, there is a little sentence at the bottom instead

### DIFF
--- a/web/massastation/src/components/SidePanels/Upload/Upload.tsx
+++ b/web/massastation/src/components/SidePanels/Upload/Upload.tsx
@@ -1,3 +1,4 @@
+import { SyntheticEvent, useRef, useState } from 'react';
 import Intl from '../../../i18n/i18n';
 
 import {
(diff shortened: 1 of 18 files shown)


web/massastation/src/mirage/server.js
```diff
@@ -177,6 +204,20 @@ function mockServer(environment = ENV.DEV) {
...
+        (schema) => {
+          let { models: accounts } = schema.accounts.all();
+
+          return accounts;
+        },
+        { timing: 500 },
+      );
+
+      this.put('websiteUploader/prepare', (schema) => {
+        return schema.create('website');
```
Thykof, line 219: this doesn't work as expected, it return {id: 4}. But it's not a problem because we don't use the response in the code

Pierregvx: I think you need to specify the other parameters

Thykof: then what is the purpose of the factory?

Pierregvx: it works for the seed but apparently not with the create

Thykof: ok, I have added a TODO comment

pivilartisant (approved):

web/massastation/src/components/SidePanels/Upload/Upload.tsx
```diff
@@ -7,8 +8,132 @@ import {
...
+  async function validate(e: SyntheticEvent): Promise<boolean> {
+    const formObject = parseForm<IFormObject>(e);
+    const { websiteName, description } = formObject;
+
+    // reset errors
+    setFileError(null);
+    setFormError(null);
+    setAccountError(null);
+
+    if (nickname === null) {
```
mazmassa, line 63: Is that work!?

```suggestion
    if (!nickname) {
```

web/massastation/src/components/SidePanels/Upload/Upload.tsx
```diff
@@ -7,8 +8,132 @@ import {
...
+    IUploadResponse
+  >('websiteUploader/prepare', {
+    'Content-Type': 'multipart/form-data',
+  });
+
+  async function validate(e: SyntheticEvent): Promise<boolean> {
+    const formObject = parseForm<IFormObject>(e);
+    const { websiteName, description } = formObject;
+
+    // reset errors
```
mazmassa, line 58:
```suggestion
```

web/massastation/src/components/SidePanels/Upload/Upload.tsx
```diff
@@ -7,8 +8,132 @@ import {
...
+    }
+
+    if (!validateWebsiteName(websiteName)) {
+      setFormError({
+        websiteName: Intl.t('search.errors.invalid-website-name'),
+      });
+      return false;
+    }
+
+    if (description === '') {
```
mazmassa, line 80:
```suggestion
    if (!description) {
```

web/massastation/src/components/SidePanels/Upload/Upload.tsx
```diff
@@ -7,8 +8,132 @@ import {
...
+
+  function handleSubmit(e: SyntheticEvent) {
+    e.preventDefault();
+
+    validate(e).then((isValid) => {
+      if (isValid) {
+        uploadWebsite(e);
+      }
+    });
+  }
```
mazmassa, line 135: Can we translate in async/await instead of then... 🤔 

```suggestion
  async function handleSubmit(e: SyntheticEvent) {
    e.preventDefault();

    let isValid = await validate(e)
    if (!isValid) return;
    uploadWebsite(e);
   }
```

web/massastation/src/custom/api/usePost.ts
```diff
@@ -1,34 +1,38 @@
...
+ * @returns
+ */
+export function usePost<TBody, TResponse = null>(
   resource: string,
 ): UseMutationResult<
-  P,
+  TResponse,
   unknown,
-  { params?: Record<string, string>; payload?: T },
+  { params?: Record<string, string>; payload?: TBody },
```
mazmassa, line 15: I am not against those changes. But we'll end up with a format for usePost/usePut here that is not the same as in the wallet. It could be tricky. 
It's a non blocking but we MUST take care of that otherwise it will be messy. @Pierregvx pinging you here to don't forget too.

web/massastation/src/mirage/server.js
```diff
@@ -86,12 +88,37 @@ function mockServer(environment = ENV.DEV) {
         },
         os: 'linux',
       }),
+      account: Factory.extend({
```
mazmassa, line 91: 👏 Beautiful!

web/massastation/src/components/SidePanels/Upload/Upload.tsx
Thykof: no we can't because handleSubmit is in a onSubmit and it must be sync

web/massastation/src/custom/api/usePost.ts
Thykof: I don't really agree, I don't think it's a problem to not have a standard, I don't see it as messy like you said. We can talk about this after the release.

maxime-carabina (approved):

web/massastation/src/pages/Index/Index.tsx
```diff
@@ -30,7 +30,6 @@ export function Index() {
   const navigate = useNavigate();
   const [pluginWalletIsInstalled, setPluginWalletIsInstalled] = useState(false);
   const [urlPlugin, setUrlPlugin] = useState('');
-  const [theme]: string = useOutletContext();
```
Thykof, line 33: @mazmassa I have error _9() is not iterable on this line so I removed it, do you have any idea?

mazmassa: This will disable the dark/light trigger. But no worries if fix your issues. I'll manage that later then in another way!

PR #825, opened 2023-06-21
fix install plugin from store

Pierregvx:
context : corrected invalid url in the post method : `url?param:value?` instead of `url?param:value`
( happened with the changes on the usePost method)


mazmassa:
> context : corrected invalid url in the post method : `url?param:value?` instead of `url?param:value` ( happened with the changes on the usePost method)

We need to map all that changes and come back on the usePost ASAP. The usePost change could be tricky...

mazmassa (approved):

pivilartisant (approved):

PR #833, opened 2023-06-21
694 Fix hanlding of .exe on window during plugin installation

Thykof:


thomas-senechal (approved):

pkg/plugin/utils/utils_windows.go
```diff
@@ -0,0 +1,14 @@
+package utils
```
peterjah, line 1: no need for  `//go:build windows`  here?

pkg/plugin/utils/utils_unix.go
```diff
@@ -0,0 +1,16 @@
...
+
+package utils
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+func pluginFileName(archiveName string) string {
+	return strings.Split(archiveName, ".zip")[0]
```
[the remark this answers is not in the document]
Thykof: True! I won't have time to refactor before the release tomorrow, can I create an issue to refactor latter?

pkg/plugin/utils/utils_windows.go
thomas-senechal: Go is supposed to automagically build this file for Windows because it ends with `_windows`.

> The naming convention is described in the documentation for the go/build package. Simply put, if your source file includes the suffix, _$GOOS.go then it will only be built on that platform. All other platforms will behave as if the file is not present. The same applies for _$GOARCH.go. The two can be combined as \_$GOOS\_$GOARCH.go, but not \_$GOARCH\_$GOOS.go.

https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool

peterjah (approved):

PR #834, opened 2023-06-21
806 enable massa station to change networks easily

B-Naoufal:


api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -374,6 +374,50 @@ paths:
...
+
+
+  /network/{network}:
+    post:
+      description: Switches the application to the specified network.
+      operationId: switchNetwork
+      parameters:
+        - in: path
+          name: network
+          type: string
```
peterjah, line 405: the type should be enum. thus you can set allowed values directly here

api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -1002,6 +1046,21 @@ definitions:
       address:
         description: Event caller.
         type: string
+  NetworkManagerItem:
+    type: object
+    description: Network manager item.
+    properties:
+      actualNetwork:
```
Thykof, line 1053: maybe currentNetwork is more appropriate

int/api/network/network_manager.go
```diff
@@ -0,0 +1,51 @@
...
+type switchNetworkHandler struct {
+	networkManager *config.NetworkManager
+}
+
+func (h *switchNetworkHandler) Handle(params operations.SwitchNetworkParams) middleware.Responder {
+	err := h.networkManager.SwitchNetwork(params.Network)
+	if err != nil {
+		return operations.NewSwitchNetworkNotFound().WithPayload(
+			&models.Error{
+				Code:    "404",
```
Thykof, line 23: it's supposed to be a constant here, like const errorNetworkNotFound = "Network-001", see how we did it for other errors

int/api/network/network_manager.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/massalabs/station/api/swagger/server/restapi/operations"
+	"github.com/massalabs/station/pkg/config"
+)
+
+func NewSwitchNetworkHandler(networkManager *config.NetworkManager) operations.SwitchNetworkHandler {
+	return &switchNetworkHandler{networkManager: networkManager}
+}
+
+type switchNetworkHandler struct {
+	networkManager *config.NetworkManager
```
peterjah, line 15: please split this file in 2.
We try to have 1 go file per handlers

api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -374,6 +374,50 @@ paths:
           description: Internal Server Error - The server has encountered a situation it does not know how to handle.
           schema:
             $ref: "#/definitions/Error"
+  /network:
+    get:
```
peterjah, line 378: There is already the endpoint massa/node that returns the current node.. Maybe it should be deprecated in favor of this one ?

peterjah (changes requested):

Thykof (changes requested):

api/swagger/server/restapi/resource/swagger.yml
B-Naoufal: yeah we will deprecate the other , this returns all networks we can switch to + the current one

peterjah (approved):
Issue to  be created to:
- Remove /massa/node endpoint
- change the wallet code to use the new endpoint instead of /massa/node

Thykof (approved):

PR #843, opened 2023-06-22
fix dashboard

lilyanB:
It's a temporary fix:
When the Index page is displayed, we display the Dashboard components with PluginWallet. The dashboard is a parent component of PluginWallet. So when there's a change to the "pluginWalletIsInstalled" variable, the Dashboard doesn't update. So the PluginWallet remains in an inactive state at all times.
with a Key we force the component to be updated.

### DIFF
--- a/web/massastation/src/pages/Index/Index.tsx
+++ b/web/massastation/src/pages/Index/Index.tsx
@@ -31,6 +31,7 @@ export function Index() {
   const [pluginWalletIsInstalled, setPluginWalletIsInstalled] = useState(false);
   const [urlPlugin, setUrlPlugin] = useState('');
   const [theme]: string = useOutletContext();
+  const [refreshPlugins, setRefreshPlugins] = useState(0);
 
   const { data: plugins } = useResource<PluginHomePage[]>('plugin-manager');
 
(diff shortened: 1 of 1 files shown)


web/massastation/src/pages/Index/Index.tsx
```diff
@@ -101,6 +106,7 @@ export function Index() {
               </Button>
             </div>
             <DashboardStation
+              key={refreshPlugins}
```
pivilartisant, line 109: does the refresh plugins have an id ?

pivilartisant (approved):

Thykof (commented):
can you describe the bug you are trying to solve in the PR description please?

Thykof (changes requested):
If it's a temporary fix, please at least add a TODO comment to explain what you are doing and why

mazmassa (approved):

mazmassa:
> If it's a temporary fix, please at least add a TODO comment to explain what you are doing and why

Sure! Indeed!

In any case the moment we rely on the Store this won't be necessary anymore!

Thykof (changes requested):
On home page, I click on Install, I have error 400 Error: given source ? is not a valid URL (parse "?": invalid URI for request)

Thykof (approved):

PR #846, opened 2023-06-22
Order Plugin List alphabetically

pivilartisant:
we order alphabetically our plugins list

### DIFF
--- a/web/massastation/src/pages/Store/StationSection/StationSection.tsx
+++ b/web/massastation/src/pages/Store/StationSection/StationSection.tsx
@@ -1,6 +1,9 @@
 import { useEffect } from 'react';
 import { useNavigate } from 'react-router-dom';
+
 import { routeFor } from '../../../utils';
+import { sortPlugins } from '../../../utils/sortArray';
+
 import Intl from '../../../i18n/i18n';
 import { UseQueryResult } from '@tanstack/react-query';
 
@@ -49,7 +52,7 @@ function StationSection({
         <>
           {plugins && plugins.length ? (
             <div className="flex gap-4 flex-wrap">
-              {plugins.map((plugin, index: number) => (
+              {sortPlugins(plugins)?.map((plugin, index) => (
                 <StationPlugin
                   key={index}
                   plugin={plugin}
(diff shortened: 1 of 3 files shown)


web/massastation/src/pages/Store/StationSection/StationSection.tsx
```diff
@@ -39,6 +39,17 @@ function StationSection({
...
+  // Sorting the plugins array alphabetically by plugin name
+  const sortedPlugins = plugins?.sort((a, b) => {
+    if (a.name < b.name) {
+      return -1;
+    }
+    if (a.name > b.name) {
+      return 1;
+    }
+    return 0;
+  });
```
mazmassa, line 51: What we are sorting here!?
Object, array of strings or numbers!? 🤔 

Isn't that enought?
```ts
my-object.sort((a, b) => a.age - b.age);
```

pivilartisant: an array of ojbects

pivilartisant: > What we are sorting here!? Object, array of strings or numbers!? thinking
> 
> Isn't that enought?
> 
> ```ts
> my-object.sort((a, b) => a.age - b.age);
> ```

I get this error with this suggestion, I do not think it is worth converting type to achieve this

The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.ts(2362)

Pierregvx: what if you parse int ?

pivilartisant: we found  a more efficient way :)

web/massastation/src/pages/Store/StoreSection/StoreSection.tsx
```diff
@@ -47,6 +47,8 @@ function StoreSection({
     );
   };
 
+  const sortedPlugins = plugins?.sort((a, b) => a.name.localeCompare(b.name));
```
mazmassa, line 50: Last: can we have that in a classical function?

mazmassa (approved):

web/massastation/src/pages/Store/StoreSection/StoreSection.tsx
pivilartisant: ofc

web/massastation/src/pages/Store/StationSection/StationSection.tsx
```diff
@@ -49,7 +52,7 @@ function StationSection({
         <>
           {plugins && plugins.length ? (
             <div className="flex gap-4 flex-wrap">
-              {plugins.map((plugin, index: number) => (
+              {sortPlugins(plugins)?.map((plugin, index) => (
```
Thykof, line 55: Are you sure you need the "?"

PR #853, opened 2023-06-22
fix theme images

mazmassa:
Since the last update dark/light mode switch are not working due to context issues. This MR fix that.


Thykof (commented):
we used to put the theme in localstorage for the wallet, why putting it in the store in massastation?

mazmassa:
> used

It still in the storage in fact. Index set that on storage to be able to persist. We are copying in the store to be able to recover on the component that needs to be updated when that change.

mazmassa:
> we used to put the theme in localstorage for the wallet, why putting it in the store in massastation?

It still in the storage in fact. Index set that on storage to be able to persist. We are copying in the store to be able to recover on the component that needs to be updated when that change.

Pierregvx (approved):

Thykof (approved):

PR #857, opened 2023-06-22
fix link Launch button

Thykof:
this will fix:

> 
> When the wallet is installed, and I click on "Launch", I'm redirected to the wallet page. --> KO
> Expected: I want the wallet to be opened in another tab (it is another app so it must open in another tab, contrary to the search or store pages that are part of the same app ;) )

from https://github.com/massalabs/station/issues/822#issuecomment-1603116073

### DIFF
--- a/web/massastation/src/pages/Index/Index.tsx
+++ b/web/massastation/src/pages/Index/Index.tsx
@@ -131,7 +131,10 @@ export function Index() {
                   iconActive={<WalletActive />}
                   iconInactive={<WalletInactive />}
                   onClickActive={() =>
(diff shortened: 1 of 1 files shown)


Pierregvx (approved):

lilyanB (changes requested):
The correct URL is:
```https://station.massa/plugin/massalabs/wallet/web-app/index```

mazmassa (approved):

Pierregvx:
> The correct URL is: `https://station.massa/plugin/massalabs/wallet/web-app/index`

we're now setting the author as Massa Labs witch is massa-labs when sluggified

lilyanB (approved):

PR #859, opened 2023-06-23
851 add network configuration file and refactor to use NetworkManager

B-Naoufal:
FLag for node server will be deprecated and remove just after the review.


pkg/config/network.go
```diff
@@ -1,71 +1,165 @@
...
-		return dnsFlag
+func (n *NetworkManager) SetNetwork(config AppConfig) {
+	n.networkMutex.Lock()
+	defer n.networkMutex.Unlock()
+
+	n.appConfig = config
+}
+
+func (n *NetworkManager) Network() *AppConfig {
+	n.networkMutex.RLock()
```
peterjah, line 122: mutex is not needed here!

peterjah (changes requested):

peterjah:
names of structures are not well chosen. You create a networkManager who contains an AppConfig.
I would have expected the opposite. an ConfigManager who contains a networkConfig, and maybe other configs

pkg/config/network.go
```diff
@@ -1,71 +1,165 @@
...
+	}
+
+	return &options
 }
 
-func GetNodeURL(urlOrNetwork string) string {
-	switch urlOrNetwork {
-	case "TESTNET":
-		return testnetNodeURL
+func (n *NetworkManager) NetworkFromString(optionStr string) (*AppConfig, error) {
```
Thykof, line 91: argument optionStr is badly named, should it be networkName?

pkg/config/network.go
```diff
@@ -1,71 +1,165 @@
...
+	networkManager.SetNetwork(appConfig)
+
+	return networkManager, nil
+}
+
+func (n *NetworkManager) SetNetworks(networks map[string]NetworkConfig) {
+	n.knownNetworks = networks
+}
+
+func (n *NetworkManager) Networks() *[]string {
```
Thykof, line 81: maybe this function should be name KnownNetworksName?

Thykof (changes requested):
please run task fmt to format and lint the code.

B-Naoufal:
> AppConfig

since massaSTATION use AppConfig everywhere , i prefer to open a seperate PR for only renaming the configmanager and appConfig an keep the diffs in this PR clean 
image
@peterjah what do you think ??

pkg/config/network.go
B-Naoufal: i think its better to have short names and the comment is very explicit here : 	// Networks retrieves a pointer to a slice of known networks from the NetworkManager.
	// It returns a pointer to a slice containing the names of the known networks.
	// The slice will be updated if the known networks are modified.
	Networks() *[]string

pkg/config/network.go
```diff
@@ -1,71 +1,168 @@
...
+	knownNetworks map[string]NetworkConfig // Known network configurations
+	networkMutex  sync.RWMutex             // Mutex for concurrent access to network configuration
+}
 
-	buildnetNodeURL    = "https://buildernet.massa.net/api/v2"
-	buildnetDNSAddress = "AS12aGWkBorEM2EpKeNyigSkoCqwdxm872g5KkzUiox3v4VosFW3F"
+// Verify at compilation time that NetworkManager implements NetworkManagerInterface.
+//
+//nolint:exhaustruct
+var _ NetworkManagerInterface = &NetworkManager{}
```
peterjah, line 26: why do you need an interface ? :/ from my experience, we use it when we need to mock the object. but here i don't see the purpose

pkg/config/interface.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// Network returns the current network configuration.
+	// It returns the network configuration represented by an AppConfig struct.
+	Network() *AppConfig
+
+	// SwitchNetwork switches the current network configuration to the specified network.
+	// selectedNetworkStr: The name of the network configuration to switch to.
+	// Returns any error encountered during the switch operation.
+	SwitchNetwork(selectedNetworkStr string) error
+}
```
peterjah, line 44: I don't think you need an interface

pkg/config/network.go
peterjah: i unresolve this because you did not removed the mutex

peterjah (changes requested):

peterjah:
> > AppConfig
> 
> since massaSTATION use AppConfig everywhere , i prefer to open a seperate PR for only renaming the configmanager and appConfig an keep the diffs in this PR clean @peterjah what do you think ??

Ok. but imho you are taking the problem in the wrong way.
you should keep AppCOnfig. but rename what you added from "networkManager" to "configManager" to be more generic..

Thykof (commented):
OK for me if Pierre's comment are solved

peterjah (approved):

Thykof (approved):

PR #865, opened 2023-06-23
fix wording in test

Thykof:


Pierregvx (approved):

Pierregvx:
could we fetch it in the repo so we don't have to do the changes 2 times ?

Thykof:
> could we fetch it in the repo so we don't have to do the changes 2 times ?

I have created issue https://github.com/massalabs/station/issues/866

PR #872, opened 2023-06-23
fix plugin request

lilyanB:
fix the name of the wallet plugin
resolve post request to run the install

### DIFF
--- a/web/massastation/src/pages/Index/Index.tsx
+++ b/web/massastation/src/pages/Index/Index.tsx
@@ -39,18 +39,17 @@ export function Index() {
   const { data: availablePlugins } =
     useResource<PluginStoreItemRequest[]>('plugin-store');
 
-  const { mutate, isSuccess, isError } = usePost<any, any>(
-    `plugin-manager?source=${urlPlugin}`,
-  );
+  const { mutate, isSuccess, isError } = usePost<null>('plugin-manager');
+  const walletName = 'Massa Wallet';
 
   useEffect(() => {
     const isWalletInstalled = plugins?.some(
-      (plugin: PluginHomePage) => plugin.name === 'wallet',
+      (plugin: PluginHomePage) => plugin.name === walletName,
     );
     setPluginWalletIsInstalled(Boolean(isWalletInstalled));
     if (!isWalletInstalled && availablePlugins) {
       const walletPlugin = availablePlugins.find(
-        (plugin: PluginStoreItemRequest) => plugin.name === 'MassaWallet',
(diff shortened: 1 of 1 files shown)


web/massastation/src/pages/Index/Index.tsx
```diff
@@ -39,18 +39,17 @@ export function Index() {
   const { data: availablePlugins } =
     useResource<PluginStoreItemRequest[]>('plugin-store');
 
-  const { mutate, isSuccess, isError } = usePost<any, any>(
-    `plugin-manager?source=${urlPlugin}`,
-  );
+  const { mutate, isSuccess, isError } = usePost<any, any>(`plugin-manager`);
```
Pierregvx, line 42: can you specify the type ?

Pierregvx (changes requested):

web/massastation/src/pages/Index/Index.tsx
Thykof, line 42:
```suggestion
  const { mutate, isSuccess, isError } = usePost<any, any>('plugin-manager');
```

Thykof (changes requested):

web/massastation/src/pages/Index/Index.tsx
lilyanB: done

Pierregvx (approved):

web/massastation/src/pages/Index/Index.tsx
```diff
@@ -39,18 +39,17 @@ export function Index() {
   const { data: availablePlugins } =
     useResource<PluginStoreItemRequest[]>('plugin-store');
 
-  const { mutate, isSuccess, isError } = usePost<any, any>(
-    `plugin-manager?source=${urlPlugin}`,
-  );
+  const { mutate, isSuccess, isError } = usePost<any, any>('plugin-manager');
```
Thykof, line 42:
```suggestion
  const { mutate, isSuccess, isError } = usePost<null>('plugin-manager');
```

web/massastation/src/pages/Index/Index.tsx
Thykof: good catch!

lilyanB: I don't know the type. The request doesn't return anything except a status.

web/massastation/src/pages/Index/Index.tsx
lilyanB: done

Thykof (approved):

PR #883, opened 2023-06-26
Fix install wallet from homepage

pivilartisant:


Thykof:
this solution use the ugly workaround

Thykof:
we chose this fix: https://github.com/massalabs/station/pull/888

PR #888, opened 2023-06-26
create parent

Pierregvx:


mazmassa (approved):

pivilartisant (approved):

web/massastation/src/pages/Index/Index.tsx
```diff
@@ -22,19 +22,39 @@ import {
...
+import { UseQueryResult } from '@tanstack/react-query';
 
 export function Index() {
+  const plugins = useResource<IMassaPlugin[]>('plugin-manager');
+  const store = useResource<IMassaStore[]>('plugin-store');
+  const { isLoading: isPluginsLoading } = plugins;
+  const { isLoading: isStoreLoading } = store;
+
+  return isPluginsLoading || isStoreLoading ? (
+    <>Loading</>
```
Thykof, line 38: could you put this in translation keys?

Pierregvx: should I use the I18n on this pr  before checking out the loading with the skeleton on another one ?

mazmassa: ...or if we don't use just set `null`

mazmassa (approved):

web/massastation/src/pages/Index/Index.tsx
Thykof: another PR

PR #893, opened 2023-06-27
add install animation

Pierregvx:
[Screencast from 28-06-2023 11:35:16.webm](https://github.com/massalabs/station/assets/90157528/9f0f33f4-e5bc-42f0-85df-8eb6dd24fd1f)

we can't see well, but the icon will pulse like this :  see [here](https://tailwindcss.com/docs/animation#pulse) 
author @pivilartisant

PR #894, opened 2023-06-27
add spin in update loading

Pierregvx:
[Screencast from 28-06-2023 12:18:49.webm](https://github.com/massalabs/station/assets/90157528/a414b65c-61cf-4473-8c4b-9e91b4ebdeb5)

author @pivilartisant

### DIFF
--- a/web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
+++ b/web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useState } from 'react';
 
-import { Button, Certificate, Plugin } from '@massalabs/react-ui-kit';
+import { Button, Certificate, Plugin, Spinner } from '@massalabs/react-ui-kit';
 import { FiArrowUpRight, FiRefreshCcw, FiTrash2 } from 'react-icons/fi';
 import {
   massalabsNomination,
@@ -36,9 +36,11 @@ export function StationPlugin({
     isRefetching,
   } = useResource<IMassaPlugin>(`plugin-manager/${id}`);
 
-  const { mutate, isSuccess } = usePost<PluginExecuteRequest>(
-    `plugin-manager/${id}/execute`,
-  );
+  const {
+    mutate,
+    isSuccess: isExecuteSuccess,
+    isLoading: isExecuteLoading,
+  } = usePost<PluginExecuteRequest>(`plugin-manager/${id}/execute`);
 
   const { mutate: deletePlugin, isSuccess: deleteSuccess } = useDelete(
     `plugin-manager/${id}`,
@@ -47,10 +49,10 @@ export function StationPlugin({
   const logoURL = `${baseAPI}/plugin-manager/${id}/logo`;
 
   useEffect(() => {
-    if (isSuccess) {
+    if (isExecuteSuccess) {
       refetch();
     }
-  }, [isSuccess]);
+  }, [isExecuteSuccess]);
 
   useEffect(() => {
     if (!isRefetching && newPlugin) {
@@ -80,9 +82,11 @@ export function StationPlugin({
     subtitleIcon: massalabsNomination.includes(author) ? <Certificate /> : null,
     content: [
       updatable && (
-        <Button variant="icon">
+        <Button variant="icon" disabled={isExecuteLoading}>
           <FiRefreshCcw
-            className="text-s-warning"
+            className={`text-s-warning${
+              isExecuteLoading ? ' animate-spin ' : ''
+            }`}
             onClick={() => updatePluginState(PLUGIN_UPDATE)}
           />
         </Button>
--- a/web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
+++ b/web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { SyntheticEvent, useEffect, useState } from 'react';
 
 import { Button, Certificate, Plugin } from '@massalabs/react-ui-kit';
 import { FiArrowUpRight, FiRefreshCw, FiTrash2 } from 'react-icons/fi';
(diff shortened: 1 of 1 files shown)


pivilartisant (commented):
little video :) pls

Pierregvx:
> little video :) pls

comming ...

web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
```diff
@@ -80,9 +82,11 @@ export function StationPlugin({
     subtitleIcon: massalabsNomination.includes(author) ? <Certificate /> : null,
     content: [
       updatable && (
-        <Button variant="icon">
+        <Button variant="icon" disabled={isExecuteLoading}>
           <FiRefreshCcw
-            className="text-s-warning"
+            className={`text-s-warning${
+              isExecuteLoading ? ' animate-spin ' : ''
```
pivilartisant, line 88: so this is a tailwind preset, unfortunately it only spins clock wise, when we need to spin our icon counter clock wise. The result is weird. We have two option : flip the icon or add a custom tailwind class se [here](https://github.com/tailwindlabs/tailwindcss/discussions/2220#discussioncomment-596742)  @Pierregvx  @mazmassa wdyt ?

Pierregvx: [Screencast from 27-06-2023 11:29:34.webm](https://github.com/massalabs/station/assets/83134981/3c326d28-e73b-4c21-954e-1604bbc6c5ad)
better ?

pivilartisant:
perfect ! does it spin for the whole update duration ?

web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
pivilartisant: perfect ! does it spin for the whole update duration ?

Pierregvx: yes sir

mazmassa: Are you disabling the button too!?

Pierregvx: it should be disabled but there is a problem with it

Pierregvx:
[Screencast from 27-06-2023 14:17:56.webm](https://github.com/massalabs/station/assets/83134981/bbff1b88-6a7c-4c0d-8a66-a3e11838595a)

web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
Pierregvx: fixed

pivilartisant (approved):
update the description with the new video

mazmassa (approved):

web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
```diff
@@ -80,10 +87,15 @@ export function StationPlugin({
...
-            className="text-s-warning"
-            onClick={() => updatePluginState(PLUGIN_UPDATE)}
+        <Button
+          variant="icon"
+          disabled={isExecuteLoading}
+          onClick={(e) => updatePluginState(e, PLUGIN_UPDATE)}
+        >
+          <FiRefreshCw
+            className={`text-s-warning${
+              isExecuteLoading ? ' animate-spin ' : ''
```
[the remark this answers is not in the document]
mazmassa: If is only the tiny spinner here maybe no, but if is a more expressive component or a page we should use an specific file/component for it.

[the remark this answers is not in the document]
mazmassa: ... or you just do as you did it here https://github.com/massalabs/station/pull/893#discussion_r1244014403! It would be nice!

pivilartisant:
@SlnPons can you Pre-QA this ? The spinner didn't seem to be the best option, so I suggested rotating the icon on click.  If you don't like it we can discuss alternatives :)

mazmassa:
The spinner didn't seem to be the best option, so I suggested rotating the icon on click. If you don't like it we can discuss alternatives :)

> The spinner didn't seem to be the best option...

Just to understand: what's wrong with the spinner? Could you, please, give a bit of context in the issue you found using it!? I would say we should keep it for loading states everywhere...

pivilartisant:
> The spinner didn't seem to be the best option, so I suggested rotating the icon on click. If you don't like it we can discuss alternatives :)
> 
> > The spinner didn't seem to be the best option...
> 
> Just to understand: what's wrong with the spinner? Could you, please, give a bit of context in the issue you found using it!? I would say we should keep it for loading states everywhere...

well there are two reasons : 
1) The spinner needs modifications in the ui to fit the bg https://github.com/massalabs/station/pull/893#issuecomment-1609061245
2)I think it's better for the user to see the update icon rotate on update rather that the icon disappear and a spinner to replace it. The spinner makes sense in the buttons because there is no icon to replace and it signals a change, but here there is already and icon that seem to fit well with a rotation

mazmassa:
> > The spinner didn't seem to be the best option, so I suggested rotating the icon on click. If you don't like it we can discuss alternatives :)
> > > The spinner didn't seem to be the best option...
> > 
> > 
> > Just to understand: what's wrong with the spinner? Could you, please, give a bit of context in the issue you found using it!? I would say we should keep it for loading states everywhere...
> 
> well there are two reasons :
> 
> 1. The spinner needs modifications in the ui to fit the bg [add a spinner during install #893 (comment)](https://github.com/massalabs/station/pull/893#issuecomment-1609061245)

Ok! In any case we'll need it with other bg too.

> 2. I think it's better for the user to see the update icon rotate on update rather that the icon disappear and a spinner to replace it. The spinner makes sense in the buttons because there is no icon to replace and it signals a change, but here there is already and icon that seem to fit well with a rotation

I see! Fair enough...

web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
```diff
@@ -64,29 +66,46 @@ export function StationPlugin({
...
     const payload = { command } as PluginExecuteRequest;
-    mutate({ payload });
+    mutateExecute({ payload });
+  }
+
+  function UpdateLoading() {
+    return (
+      <>
+        <FiRefreshCw className={`text-s-warning animate-spin`} />
+      </>
```
mazmassa, line 80:
```suggestion
        <FiRefreshCw className={`text-s-warning animate-spin`} />
```

mazmassa (approved):

PR #898, opened 2023-06-27
web: migrate mirage from JS to TS

maxime-carabina:
We are migrating mirage from JS to TS.


mazmassa (approved):

web/shared/interfaces/IPlugin.ts
Thykof, line 1: do we use it in web/massastation?

Thykof (approved):

web/shared/interfaces/IPlugin.ts
maxime-carabina: Yes

Thykof: okay, we will create a task to remove web/home and web/node_manager, so if we only use this file in web/massastation, we can move it somewhere in web/massastation (no priority though)

PR #900, opened 2023-06-28
fix usePost

pivilartisant:
Fix use post where we get no params

### DIFF
--- a/web/massastation/src/custom/api/usePost.ts
+++ b/web/massastation/src/custom/api/usePost.ts
@@ -16,7 +16,7 @@ export function usePost<TBody, TResponse = null>(
   unknown
 > {
   const baseURL = import.meta.env.VITE_BASE_API;
-  const url = `${baseURL}/${resource}`;
+  let url = `${baseURL}/${resource}`;
 
   return useMutation<
     TResponse,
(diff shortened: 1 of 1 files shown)


Pierregvx (approved):
LGTM

Thykof (approved):

web/massastation/src/custom/api/usePost.ts
```diff
@@ -28,9 +28,10 @@ export function usePost<TBody, TResponse = null>(
     mutationFn: async ({ params, payload }) => {
       const queryParams = new URLSearchParams(params).toString();
       if (queryParams) {
-        url.concat(`?${queryParams}`);
+        url = `${url}?${queryParams}`;
```
maxime-carabina, line 31: `.concat` is better than doing this 🤔

Pierregvx: it didn't work with xoncat

pivilartisant: this does not work but if you want we can do this : 
```
url = url.concat(`?${queryParams}`);
```

Thykof: I think that `    url = `${url}?${queryParams}`;` is easier to read that concat

maxime-carabina: > this does not work but if you want we can do this :
> 
> ```
> url = url.concat(`?${queryParams}`);
> ```

Thank you! Better on this way

PR #901, opened 2023-06-28
macos postinstall: change root ca permissions

Thykof:


gregLibert (approved):

thomas-senechal (approved):

Thykof:
Thanks for your reviews, before merge I want to download and test the .pkg file. I'll keep you posted

Thykof:
could you please have a look again, I have change because it wasn't working. Now I tested the .pkg and it works with this solution.

PR #903, opened 2023-06-28
🔥 remove legacy frontends

Thykof:


pivilartisant:
"-20,873" ahha

pivilartisant (approved):

B-Naoufal:
The only issue is we where able to edit a website or re upload it using the old front end ! I am not sure we have the same functionality with the new one !? @Thykof ( this functionality was supported in /websiteuploader page

Thykof:
> The only issue is we where able to edit a website or re upload it using the old front end ! I am not sure we have the same functionality with the new one !? @Thykof ( this functionality was supported in /websiteuploader page

You are right, @SlnPons should we keep the old web-on-chain frontend because the new front isn't iso with the old one?

Thykof:
I use sleep 5 to passes the frontend tests, it's not optimal, I'm looking for a better solution

api/test/robot_tests/front_ends/front_ends.robot
```diff
@@ -13,86 +13,13 @@ Suite Teardown      Close All Browsers
...
-GET /store/{resource} with invalid resource
-    ${response}=    GET    ${API_URL}/store/invalid    expected_status=${STATUS_NOT_FOUND}
-
-GET /search/{resource} with invalid resource
-    ${response}=    GET    ${API_URL}/search/invalid    expected_status=${STATUS_NOT_FOUND}
-
-GET /websiteUploader/{resource} with invalid resource
-    ${response}=    GET    ${API_URL}/websiteUploader/invalid    expected_status=${STATUS_NOT_FOUND}
+GET /web/index
+    Open Massa Station Page    ${API_URL}/web/index
```
thomas-senechal, line 24: Shouldn't that be :
```suggestion
    Open Massa Station Page    ${API_URL}/web/index.html
```

api/test/robot_tests/web_on_chain/web_on_chain.robot
```diff
@@ -23,8 +23,11 @@ PUT /websiteUploader/prepare
     ...    ${API_URL}/websiteUploader/prepare
     ...    data=${data}
     ...    files=${file}
-    ...    expected_status=${STATUS_OK}
+    ...    expected_status=any
```
thomas-senechal, line 26: Shouldn't that be revert to check for `STATUS_OK` ?

SlnPons:
> > The only issue is we where able to edit a website or re upload it using the old front end ! I am not sure we have the same functionality with the new one !? @Thykof ( this functionality was supported in /websiteuploader page
> 
> You are right, @SlnPons should we keep the old web-on-chain frontend because the new front isn't iso with the old one?

I think that only the new front-end will be used to upload websites. HOWEVER do you know if all "features" available in the old front-end are also available using CLI? 
--> the rational here is to enable developers to upload a website with all the functionalities available (because that would be stupid to hide this) but using CLI and not front-end. 
Not sure to be very clear ... let me know.

thomas-senechal (commented):
A few comments

I'm also wondering if removing the old front end for website upload since we lost some functionalities such as edit and the list of your websites

api/test/robot_tests/front_ends/front_ends.robot
Thykof: no, react router take the lead on the path

api/test/robot_tests/web_on_chain/web_on_chain.robot
Thykof: no, I assert the status code bellow line 30

thomas-senechal:
> I am not aware of a web-on-chain CLI tool up to date for buildnet

Isn't really a "CLI tool", it's just a `curl` command ^^

thomas-senechal (approved):

Thykof:
We have written an internal documentation about how to use webonchain API here https://www.notion.so/massa-innoteam/Use-web-on-chain-with-CLI-5d8d47acbaa8466daac52e47e1c57b41

thomas-senechal (approved):

PR #911, opened 2023-06-28
Fix Store useDelete && GET 404

pivilartisant:


web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
```diff
@@ -135,7 +137,11 @@ export function StationPlugin({
       </Button>,
     ],
   };
-  return <Plugin {...(status === PluginStatus.Up ? argsOn : argsOff)} />;
+  return isLoading || isRefetching ? (
+    <div> whoop whiooop </div>
```
Pierregvx, line 141: remove the whoop whoop haha

Pierregvx (changes requested):

web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
pivilartisant: i was caught

web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
```diff
@@ -53,9 +54,10 @@ export function StationPlugin({
       refetch();
     }
   }, [isExecuteSuccess]);
-
+  // add isLoading conditional so that the useRessource isn't called b4
```
Thykof, line 57: useDelete can't call useRessource, this is not the source of the issue, maybe it solve it though.

useRessource is called because react query consider it should refresh the data.

pivilartisant: ah I see what you mean, would you recommend going forward with this fix or does it need a more in depth investigation & fix ?

Thykof: probably not depth investigation if your fix work, but the comment is misleading, either rewrite it or remove it

pivilartisant: so i did some more testing, indeed the fix is not perfect, on slow 3G for example the useDelete is never called.... but I have the feeling that for right now that is out of my scope

Thykof (approved):

web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
Thykof: yes, we will merge this PR and reopen the linked issue to be priorized

Pierregvx (approved):

PR #931, opened 2023-06-30
ubuntu post install script: waiting for network manager

Thykof:


installer/deb/scripts/postinst
```diff
@@ -59,6 +59,10 @@ set_local_dns() {
 
 # Download and install mkcert, generate a certificate for MassaStation.
 generate_certificate() {
+    while ! ifconfig | grep "192.168." > /dev/null; do
+        echo "waiting for network manager to be up and running"
+        sleep 2
```
Thykof, line 64: WARN: this is a potential infinite loop

Thykof: chat gpt suggestion:

```
#!/bin/bash

counter=0
timeout=10  # Timeout in seconds

while ! ifconfig | grep "192.168." > /dev/null && ((counter < timeout)); do
    echo "Waiting for network manager to be up and running"
    sleep 1
    ((counter+=1))
done

if ((counter >= timeout)); then
    echo "Timeout reached. Network manager is not available."
else
    echo "Network manager is up and running."
fi
```

thomas-senechal: I think you should create a `wait_for_network` function that contains what ChatGPT gave you, and call it at the begining of `generate_certificate`

Thykof: yep!

installer/deb/scripts/postinst
```diff
@@ -57,8 +57,26 @@ set_local_dns() {
...
+    timeout=20  # Timeout in seconds
+
+    while ! ifconfig | grep "192.168." > /dev/null && ((counter < timeout)); do
+        echo "Waiting for network manager to be up and running"
+        sleep 1
+        ((counter+=1))
+    done
+
+    if ((counter >= timeout)); then
+        echo "Timeout reached. Network manager is not available."
```
thomas-senechal, line 71: You should use `fatal` to print the error and stop the installation since there is a network issue

```suggestion
        fatal "Network manager is not available: Timeout reached. "
```

installer/deb/scripts/postinst
```diff
@@ -57,8 +57,26 @@ set_local_dns() {
...
+    else
+        echo "Network manager is up and running."
```
thomas-senechal, line 73: I'm not sure it is really useful to print this 🤔 
```suggestion
```

thomas-senechal (approved):

Pierregvx (changes requested):
got 
`/var/lib/dpkg/info/massastation.postinst: line 64: ifconfig: command not found`
=> add ifconfig package

Thykof:
I tested on ubuntu 22, it works with these artifacts https://github.com/massalabs/station/actions/runs/5423759247?pr=931

Pierregvx (approved):

Thykof:
curl package is also missing as pointed out by @Pierregvx , @thomas-senechal should we add curl package in the dependency list?

thomas-senechal (approved):

installer/deb/create_deb.sh
```diff
@@ -86,6 +86,7 @@ Maintainer: Massa Labs <massa.net>
 Homepage: https://github.com/massalabs/station
 Description: An entrance to the Massa blockchain.
     MassaStation is a secured gateway to the Massa blockchain. This application provides a user-friendly way to access, use and build on the Massa blockchain while keeping you safe from the dangers of the internet.
+Depends: curl iproute2
```
Pierregvx, line 89:
```suggestion
Depends: curl,iproute2
```

Pierregvx (changes requested):

Pierregvx (approved):

Thykof:
I tested on ubuntu 22

PR #932, opened 2023-06-30
create logs folder in windows installer

Pierregvx:
image

### DIFF
--- a/pkg/config/logger.go
+++ b/pkg/config/logger.go
@@ -18,8 +18,8 @@ func NewLogger() *zap.SugaredLogger {
 		log.Fatal(err)
 	}
 
-	logFilePath := filepath.Join(logDir, "logs", "massastation.log")
-	logDirPath := filepath.Dir(logFilePath)
+	logDirPath := filepath.Join(logDir, "logs")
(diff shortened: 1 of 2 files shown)


pkg/config/logger.go
```diff
@@ -18,11 +18,14 @@ func NewLogger() *zap.SugaredLogger {
 		log.Fatal(err)
 	}
 
-	logFilePath := filepath.Join(logDir, "massastation.log")
+	logFilePath := filepath.Join(logDir, "logs", "massastation.log")
+	logDirPath := filepath.Dir(logFilePath)
```
Thykof, line 22: I feel it's more readable like this:
```suggestion
	logDirPath := filepath.Join(logDir, "logs")
	logFilePath := filepath.Join(logDirPath, "massastation.log")
```

Thykof (approved):

pkg/config/logger.go
Pierregvx: done

Thykof:
I tested, it works on windows and Macos arm

thomas-senechal (approved):

PR #935, opened 2023-06-30
refactor utils package

Thykof:


Thykof:
I tested on linux and windows with https://github.com/massalabs/station/actions/runs/5423389052?pr=935 and it works

PR #937, opened 2023-06-30
web: add absolute relative path

maxime-carabina:
We are adding absolute path.

Like this : `@/pages` or `@/const`


Thykof (approved):
great! sometime we let some ../..., is it intended? stuff about shared folder is legacy, we should not waste time with this. I opened a draft PR to get rid of it https://github.com/massalabs/station/pull/903

maxime-carabina:
> great! sometime we let some ../..., is it intended? stuff about shared folder is legacy, we should not waste time with this. I opened a draft PR to get rid of it #903

Oh sorry! If it's going to be removed, would you prefer I leave it as it is for now and we'll make the change in the other PR? @Thykof

Thykof:
> > great! sometime we let some ../..., is it intended? stuff about shared folder is legacy, we should not waste time with this. I opened a draft PR to get rid of it #903
> 
> Oh sorry! If it's going to be removed, would you prefer I leave it as it is for now and we'll make the change in the other PR? @Thykof

no let's move on with your PR, mine is blocked.

pivilartisant (approved):

PR #938, opened 2023-07-03
forbid second registering

Pierregvx:


int/api/myplugin/plugin.go
```diff
@@ -31,8 +31,9 @@ const (
 
 	errorCodePluginInstallationInvalidSource = "Plugin-0010"
 
-	errorCodePluginRegisterUnknown     = "Plugin-0020"
-	errorCodePluginRegisterInvalidData = "Plugin-0020"
+	errorCodePluginRegisterUnknown           = "Plugin-0020"
+	errorCodePluginRegisterInvalidData       = "Plugin-0020"
+	errorCodePluginRegisterAlreadyRegistered = "Plugin-0020"
```
thomas-senechal, line 36: Those error should have different identifiers since it should allow us to handle the error and print the right error message in the front end.
```suggestion
	errorCodePluginRegisterUnknown           = "Plugin-0020"
	errorCodePluginRegisterInvalidData       = "Plugin-0021"
	errorCodePluginRegisterAlreadyRegistered = "Plugin-0022"
```

pivilartisant (approved):

thomas-senechal (changes requested):
CI is failing

int/api/myplugin/execute.go
```diff
@@ -25,35 +24,37 @@ func (e *execute) Handle(params operations.PluginManagerExecuteCommandParams) mi
 
 	config.Logger.Debugf("[POST /plugin-manager/%s/execute] command: %s", params.ID, cmd)
 
-	plugin, err := e.manager.Plugin(params.ID)
+	plgn, err := e.manager.Plugin(params.ID)
```
thomas-senechal, line 27: I don't really like removing letters from a word in variable names
I think you should rename it to `plugin` and import the `plugin` package as `pluginPkg` for example (in the case of the import, it's fine to remove letter from "package" because the directory is already named `pkg`)

```suggestion
	plugin, err := e.manager.Plugin(params.ID)
```

thomas-senechal (approved):
Nice job !

Only one non blocking comment

int/api/myplugin/execute.go
Pierregvx: done

PR #940, opened 2023-07-03
Use nmcli instead of ip to check for internet connectivity in debian …

thomas-senechal:
…postinst script.

### DIFF
--- a/installer/deb/scripts/postinst
+++ b/installer/deb/scripts/postinst
@@ -59,9 +59,9 @@ set_local_dns() {
 
 wait_for_network() {
     counter=0
-    timeout=30  # Timeout in seconds
+    timeout=30 # Timeout in seconds
 
-    while ! ip a | grep "192.168." > /dev/null && ((counter < timeout)); do
+    while [[ $(nmcli networking connectivity check) == "none" ]]; do
         echo "Waiting for network manager to be up and running"
         sleep 1
         ((counter+=1))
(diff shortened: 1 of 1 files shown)


installer/deb/scripts/postinst
```diff
@@ -59,9 +59,9 @@ set_local_dns() {
 
 wait_for_network() {
     counter=0
-    timeout=30  # Timeout in seconds
+    timeout=30 # Timeout in seconds
 
-    while ! ip a | grep "192.168." > /dev/null && ((counter < timeout)); do
+    while [[ $(nmcli networking connectivity check) == "none" ]]; do
```
Thykof, line 64: you remove this part `&& ((counter < timeout))`

installer/deb/scripts/postinst
```diff
@@ -78,7 +78,7 @@ generate_certificate() {
     curl -sL $MKCERT_URL -o mkcert || fatal "mkcert download failed."
     chmod +x mkcert || fatal "mkcert permission change failed."
 
-    # We ensure that CAROOT is set. Otherwise, mkcert installation will fail. 
+    # We ensure that CAROOT is set. Otherwise, mkcert installation will fail.
```
Thykof, line 81: great! do you use a .sh formatting tool?

thomas-senechal: Yes, but I don't really like it because it changes a lot of stuff in a way that I found less readable, such as removing the indentation of the switch cases
It's "shell-format" in VSCode

Thykof (approved):

B-Naoufal (approved):

SlnPons:
Should I test this? 
If yes, how? I think I would need your help here...

PR #942, opened 2023-07-03
Fix Store Calls

pivilartisant:
we fix various errors tracked here : https://github.com/massalabs/station/issues/910


web/massastation/src/pages/Store/StoreSection/StorePlugin.tsx
```diff
@@ -33,10 +34,10 @@ function StorePlugin(props: StorePluginProps) {
   const params = { source: url };
 
   useEffect(() => {
-    if (isInstallSuccess) {
+    if (isInstallSuccess && !isInstallLoading) {
```
Pierregvx, line 37: are you sure we need that ? isn't isInstallLoading always false if isInstallSuccess = true ?

mazmassa (approved):

PR #948, opened 2023-07-04
Testnet 23 update

ASAPSegfault:


Thykof:
api test are failing because it used buildnet21.

int/api/websites/all.go
```diff
@@ -19,7 +19,7 @@ import (
 const (
 	dateFormat          = "2006-01-02"
 	ownedPrefix         = "owned"
-	ownerKey            = "owner"
+	ownerKey            = "OWNER"
```
B-Naoufal, line 22: maybe we should add a comment here to refer where we declare those keys on SC level ? what do you think ?

pkg/config/config_network.yaml
```diff
@@ -1,14 +1,13 @@
 testnet:
-  DNS: AS12YMz7NjyP3aeEWcSsiC58Hba8UxHapfGv7i4PmNMS2eKfmaqqC
+  DNS: AS12MGDGVB3xNDqzGidJM5fcAbjF1ZP5XAecm5ww4RU7sfQd4cMKk
   URLs:
     - https://test.massa.net/api/v2
+  Default: true
```
B-Naoufal, line 5: are we sure that we want to keep testnet as default network ?

B-Naoufal (approved):
Well Done,  thank you !

pkg/config/config_network.yaml
Thykof: no, true I just changed it, I was wainting for the buildnet to pass to v24

peterjah (approved):

cmd/massastation/main.go
```diff
@@ -59,18 +59,19 @@ func ParseFlags() api.StartServerFlags {
 }
 
 func main() {
+	err := initialize.Logger()
+	if err != nil {
+		log.Fatalf("while initializing logger: %s", err.Error())
+	}
+
```
gregLibert, line 66: defer logger.Close() should be here

cmd/massastation/main.go
```diff
@@ -59,18 +59,19 @@ func ParseFlags() api.StartServerFlags {
...
 	flags := ParseFlags()
 	if flags.Version {
 		//nolint:forbidigo
 		fmt.Printf("Version:%s\n", config.Version)
+		logger.Close()
```
gregLibert, line 71: use less if defer is moved

int/api/websites/all.go
```diff
@@ -17,12 +17,12 @@ import (
...
 const (
-	dateFormat          = "2006-01-02"
-	ownedPrefix         = "owned"
-	ownerKey            = "owner"
-	blackListKey        = "blackList"
-	secondsToMilliCoeff = 1000
-	faviconIcon         = "favicon.ico"
+	// Keys declared in the DNS smart contract.
+	ownedPrefix  = "owned"
+	ownerKey     = "OWNER"
```
gregLibert, line 22: :see_no_evil:

pkg/convert/byteConverter.go
```diff
@@ -30,6 +30,10 @@ func U32ToBytes(nb int) (bytes []byte) {
 	return
 }
 
+func StringToBytesWithoutPrefixLength(str string) []byte {
```
gregLibert:
```suggestion
func StringToBytesWithoutLengthPrefix(str string) []byte {
```

pkg/node/sendoperation/signer/interface.go
```diff
@@ -0,0 +1,12 @@
+package signer
+
+//nolint:tagliatelle
+type SignOperationResponse struct {
+	PublicKey     string `json:"publicKey"`
+	Signature     string `json:"signature"`
+	CorrelationID string `json:"correlationId,omitempty"`
+}
+
+type Signer interface {
```
gregLibert, line 10: nickel!

pkg/node/sendoperation/signer/signer.go
```diff
@@ -0,0 +1,82 @@
...
+	httpRawResponse, err := ExecuteHTTPRequest(
+		http.MethodPost,
+		WalletPluginURL+"accounts/"+nickname+"/sign",
+		bytes.NewBuffer(operation),
+	)
+	if err != nil {
+		res := RespError{"", ""}
+		_ = json.Unmarshal(httpRawResponse, &res)
+
+		return nil, fmt.Errorf("calling executeHTTPRequest for call: %w, message: %s", err, res.Message)
```
gregLibert, line 38:
```suggestion
		return nil, fmt.Errorf("calling executeHTTPRequest: %w, message: %s", err, res.Message)
```

pkg/node/sendoperation/signer/signer.go
```diff
@@ -0,0 +1,82 @@
...
+	resp, err := HTTPClient.Do(request)
+	if err != nil {
+		return nil, fmt.Errorf("aborting during HTTP request: %w", err)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("reading request body: %w", err)
+	}
+
```
gregLibert, line 74: defer erspo.Body.Close() should be here.

cmd/massastation/main.go
Thykof: yeah I had lint issue because bellow we do `os.Exit(0)`: "exitAfterDefer: os.Exit will exit, and `defer logger.Close()` will not run"

*Side note:*
Why I move the logger initialization: because `ParseFlags` use the logger.

pkg/node/sendoperation/signer/signer.go
Thykof: why? why not just before the `io.ReadAll(resp.Body)` like https://stackoverflow.com/a/62573623/5903242 suggests?

gregLibert (approved):

PR #953, opened 2023-07-04
Set Application Bg

pivilartisant:
Status : blocked 
Why ? 
No clear way to use tailwind and set the bg to primary. It is possible to hard code it, but then it breaks the light mode


pivilartisant:
closed because there is no current plan to address this subject

PR #954, opened 2023-07-04
946 add installation of ca in system store for windows only

gregLibert:


pkg/certificate/store/store_windows.go
```diff
@@ -0,0 +1,154 @@
...
+	"fmt"
+	"syscall"
+	"unsafe"
+)
+
+const (
+	CRYPT_E_NOT_FOUND = 0x80092004
+)
+
+// The functions bellow return an error even though they succeed.
```
peterjah, line 14: These functions seems inspired from mkcert. maybe it worth to put the link here in comment

peterjah (approved):

pkg/certificate/store/store_windows.go
```diff
@@ -0,0 +1,154 @@
...
+
+		certSyscall = (*syscall.CertContext)(unsafe.Pointer(certPtr))
+
+		// Parse cert
+		certBytes := (*[1 << 20]byte)(unsafe.Pointer(certSyscall.EncodedCert))[:certSyscall.Length]
+		certX509, err := x509.ParseCertificate(certBytes)
+
+		// Ignore parsing errors
+		if err != nil || certX509.SerialNumber == nil {
+			continue
```
Pierregvx, line 140: could we add a warn logging here ?

Pierregvx (approved):

pkg/certificate/store/store_windows.go
gregLibert: What the point? What will we do with the warning ? The context here is that we are listing all the certificates of the ROOT trust store. if one of them is broken, we can do nothing...

pkg/certificate/store/store_windows.go
```diff
@@ -0,0 +1,154 @@
...
+	"crypto/x509"
+	"fmt"
+	"syscall"
+	"unsafe"
+)
+
+const (
+	CRYPT_E_NOT_FOUND = 0x80092004
+)
+
```
Thykof, line 13:
```suggestion

// inspired by https://github.com/FiloSottile/mkcert

```

pkg/certificate/store/store_windows.go
Thykof: like this https://github.com/massalabs/station/pull/954/files#r1252752318 ?

pkg/certificate/store/store_windows.go
Pierregvx: ok I see then I guess it is not useful indeed

thomas-senechal (approved):

pkg/certificate/store/store_windows.go
Thykof: we will do it in a following PR

PR #958, opened 2023-07-04
no rollback so no DNS broken

Thykof:


pivilartisant (approved):

installer/windows/build_installer.py
```diff
@@ -289,14 +289,6 @@ def create_wxs_file():
             Impersonate="no"
             Return="check"
         />
-        <CustomAction
-            Id="RollbackAcrylicInstall"
-            Directory="AcrylicDNSProxy"
-            ExeCommand="cmd /c &quot;[AcrylicDNSProxy]UninstallAcrylicService.bat&quot;"
-            Execute="rollback"
-            Impersonate="no"
-            Return="ignore"
```
thomas-senechal, line 298: IMO, we should keep this since there is a high probability that we will reuse it later

And since it has been removed from the `InstallExecuteSequence`, it won't be run

thomas-senechal (approved):
Only one non blocking comment

installer/windows/build_installer.py
Thykof: yeah but I don't like to let dead code

PR #965, opened 2023-07-06
962 add nss check for ubuntu

gregLibert:


int/config/check.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if !certCa.IsKnownByNSSDatabases() {
+		logger.Logger.Debug("the CA is not known by at least one local NSS database.")
+
+		err := certCa.AddToNSSDatabases()
+		if err != nil {
+			// non blocking error
+			logger.Logger.Warnf("failed to add the CA to NSS: %s.", err)
+			logger.Logger.Warn(caFailureConsequence)
+		}
```
Thykof, line 77:
```suggestion
		caCheckNonBlockingError("failed to add the CA to NSS", err)
```

pkg/certificate/store/nssdb_unix.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+
+	return nil
+}
+
+// Add adds the certificate to the NSS databases.
+func (n *NSSDatabases) Add(certPath string, certificateName string) error {
+	return n.executeOnPaths(func(path string) error {
+		err := runCertutilCommand("-A", "-d", path, "-t", "C,,", "-n", certificateName, "-i", certPath)
```
Thykof, line 47: `C,,` smells weird, what does it mean with two ","?

pkg/certificate/store/nssdb_unix.go
```diff
@@ -0,0 +1,125 @@
...
+		if err != nil {
+			logger.Logger.Errorf("failed to parse nssdb pattern (%s):  %v", path, err)
+
+			continue
+		}
+
+		// if the path is a pattern, we need to filter the dirctories not containing a database
+		if strings.Contains(path, "*") {
+			for _, match := range matches {
+				dbFiles, _ := filepath.Glob(filepath.Join(match, databasePattern))
```
Thykof, line 97: [optional]

you don't check the err here, golangcilint should complain.

pkg/certificate/store/nssdb_unix.go
```diff
@@ -0,0 +1,125 @@
...
+		if err != nil {
+			return false
+		}
+	}
+
+	return true
+}
+
+// filterExistingPath filters the given paths and returns only the existing ones.
+func filterExistingPath(theoricPath []string) []string {
```
Thykof, line 83: [optional, nitpicking]

This function has a high cyclomatic complexity. consider extracting the nested for loop into a separate function.

pkg/certificate/store/nssdb_windows.go
```diff
@@ -0,0 +1,28 @@
+package store
+
+import "fmt"
+
+// NSSDatabases is a struct that represents the known NSS databases on the system.
+type NSSDatabases struct {
+	Paths []string
+}
```
Thykof, line 8: this struct is already declared in nssdb_unix. Unless you anticipate stuff to be added for windows and not unix, consider moving it to a new file `nssdb.go`

pkg/certificate/store/nssdb_unix.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return nil
+}
+
+// Add adds the certificate to the NSS databases.
+func (n *NSSDatabases) Add(certPath string, certificateName string) error {
+	return n.executeOnPaths(func(path string) error {
+		err := runCertutilCommand("-A", "-d", path, "-t", "C,,", "-n", certificateName, "-i", certPath)
+		logger.Logger.Debugf("adding the certificate to the NSS database (%s): %s", path, err)
```
Thykof, line 48: Here it's logging `adding the certificate to the NSS database (/Users/nathanmassa/Library/Application Support/Firefox/Profiles/k0e9460f.default-release): %!s(<nil>)`

```suggestion
		if err != nil {
			logger.Logger.Debugf("fail adding the certificate to the NSS database (%s): %s", path, err)
		}
```

pkg/certificate/store/nssdb_unix.go
gregLibert: It means trust this server to issue server certificate for TLS website.
Source: `certituil --help -A`
Details of trust args (`-t`):
image

int/config/check.go
```diff
@@ -100,21 +112,22 @@ func checkNSS(certPath string) error {
 		return nssCheckNonBlockingError("failed to instantiate the certutil runner", err)
 	}
 
-	manager := nss.NewManager([]string{}, service, logger.Logger)
+	loggerInstance := &NSSManagerLogger{}
+	manager := nss.NewManager([]string{}, service, loggerInstance)
```
gregLibert, line 116: Why can't you pass logger here instead of loggerInstance ?

pkg/certificate/authority.go
Thykof, line 1: The PR re-add this file because it needs it in int/config/check. I haven't found a way during the update of the branch with main branch.

int/config/check.go
Thykof: because `logger` is a package now. the `nss.NewManager` require a interface that implement `Debugf` and `Errorf`. So I created a struct `NSSManagerLogger` that wrap the logger package.

I'd not be surprise if we can do better.

pkg/certificate/authority.go
Thykof: I see that here you moved the content of pkg/certificate/authority.go. So I think it's good to have it here

https://github.com/massalabs/station/pull/991/files#diff-7528617d54c1073332a471d5e01663342787a04172f9163be739d56c538aa7d4

Thykof:
@gregLibert you can review

Thykof (approved):

pkg/nss/manager.go
```diff
@@ -17,6 +30,7 @@ type Manager struct {
 }
 
 // NewManager returns a new Manager instance.
+// It will the NSS databases corresponding to the given paths.
```
Thykof, line 33: ?

```suggestion
// It will create the NSS databases corresponding to the given paths.
```

Thykof (changes requested):
sometimes to refer to CertUtilService and other times to CertUtilServicer. Is it the same think (hence a typo)?

pkg/nss/interface.go
```diff
@@ -0,0 +1,23 @@
+package nss
+
+// Logger is the interface used by the NSS package to log messages.
+type Logger interface {
+	Debugf(format string, v ...interface{})
+	Errorf(format string, v ...interface{})
+}
+
+// CertUtilServicer encapsulates operations on NSS database using certutil command.
+type CertUtilServicer interface {
```
thomas-senechal, line 10: Non blocking,

Is it a common thing to name interfaces that way ? 
IMO, `CertUtilServiceInterface` would be clearer since we directly know that it is an interface

thomas-senechal (approved):

pkg/nss/interface.go
thomas-senechal: > sometimes to refer to CertUtilService and other times to CertUtilServicer. Is it the same think (hence a typo)?

@Thykof `CertUtilServicer` is the interface, and `CertUtilService` a struct implementing the interface

WDYT about this suggestion ?

int/config/check.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if !manager.HasCA(configuration.CertificateAuthorityName) {
+		logger.Debug("the CA is not known by at least one local NSS database.")
+
+		err := manager.AddCA(configuration.CertificateAuthorityName, certPath)
+		if err != nil {
+			// non blocking error
+			logger.Warnf("failed to add the CA to NSS: %s.", err)
+			logger.Warn(caFailureConsequence)
```
gregLibert, line 126:
```suggestion
			caCheckNonBlockingError(fmt.Errorf("failed to add the CA to NSS: %w", err))
```

pkg/nss/manager.go
gregLibert: manage

int/config/check.go
```diff
@@ -0,0 +1,135 @@
...
+
+	service, err := nss.NewCertUtilService(runner)
+	if err != nil {
+		return nssCheckNonBlockingError("failed to instantiate the certutil service", err)
+	}
+
+	loggerInstance := &NSSManagerLogger{}
+	manager := nss.NewManager([]string{}, service, loggerInstance)
+
+	if !manager.HasCA(configuration.CertificateAuthorityName) {
```
gregLibert, line 119: The method 'AppendDefaultNSSDatabasePaths()' needs to be called on manager before HasCA. if not, no database will be checked.

gregLibert (commented):
@Thykof my review

pkg/nss/interface.go
Thykof: Ho I see, thanks.

IMO no we should not name a interface BlaBlaInterface, not IBlaBla for many reason I found on internet.

cmd/massastation/main.go
```diff
@@ -8,6 +8,7 @@ import (
 	"path"
 
 	"github.com/massalabs/station/int/api"
+	intconf "github.com/massalabs/station/int/config"
```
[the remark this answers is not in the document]
Thykof: yes

pkg/nss/interface.go
gregLibert: The interface are named XXXer mostly when they contain only one function:
```golang
type Writer interface {
     Write(file string, content []byte) error
}
```
Here servicer is not an english word

gregLibert:
> sometimes to refer to CertUtilService and other times to CertUtilServicer. Is it the same think (hence a typo)?

I dont think so. Otherwise the compiler would shout...

int/config/check.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	certPath := filepath.Join(caRootPath, configuration.CertificateAuthorityFileName)
+
+	keyPath := filepath.Join(caRootPath, configuration.CertificateAuthorityKeyFileName)
+
+	err = checkCertificate(certPath, keyPath)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
Thykof: and so also checkNSS?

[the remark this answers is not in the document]
gregLibert: Yes, this is also true for checknss

int/config/check.go
Thykof: https://github.com/massalabs/station/issues/1042

Thykof (approved):
You can review @gregLibert

int/config/check.go
```diff
@@ -0,0 +1,126 @@
...
+
+	err = checkNSS(certPath)
+	if err != nil {
+		return nonBlockingError("Error while checking NSS", failureConsequence, err)
+	}
+
+	return nil
+}
+
+// nonBlockingError logs a non blocking error.
```
thomas-senechal, line 40: I would add something about the fact that we always return a `nil` error such as:
```suggestion
// nonBlockingError logs a non blocking error. It always returns a `nil` error to be used in `return`.
```

int/config/check.go
```diff
@@ -0,0 +1,126 @@
...
+func nonBlockingError(context string, consequence string, err error) error {
+	if err != nil {
+		logger.Warnf("%s: %s.", context, err)
+		logger.Warn(consequence)
+	}
+
+	return nil
+}
+
+// caNonBlockingError logs a non blocking error related to the CA loading failure.
```
thomas-senechal, line 50: I would also add something about the `nil` always returned here since it uses `nonBlockingError` under the hood

pkg/nss/manager.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+)
+
+// databasePattern is the pattern used to find NSS databases with dynamic path configuration.
+const databasePattern = "cert*.db"
+
+// manager encapsulates operations to execute on all the NSS databases of the operating system.
```
thomas-senechal, line 25:
```suggestion
// Manager encapsulates operations to execute on all the NSS databases of the operating system.
```

thomas-senechal (approved):
Good job ! 

Just a few non blocking comments

int/config/check.go
Thykof: I think it's ok because thanks to the function name we understand that it will never block

gregLibert (commented):
Good job !

PR #967, opened 2023-07-06
882 logo problem on windows

pivilartisant:
This is a proposition, it is not tested. But @maxime-carabina and I think it could advance us in finding a solution.


pivilartisant:
mistake

PR #978, opened 2023-07-07
976 centralize cross package configuration in intconf package

gregLibert:


pkg/certificate/authority.go
```diff
@@ -6,14 +6,10 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/massalabs/station/int/configuration"
```
Thykof, line 9: here in pkg/certificate we are importing int/configuration so when package certificate depends on int of massa station. So we can't extract package certificate.

I had in mind we pass as an argument to CA::Load the CAPath.

int/configuration/configuration.go
```diff
@@ -0,0 +1,38 @@
...
+	switch {
+	case runtime.GOOS == "windows":
+		ex, err := os.Executable()
+		if err != nil {
+			return "", fmt.Errorf("getting executable path: %w", err)
+		}
+
+		return filepath.Dir(ex), nil
+
+	case runtime.GOOS == "linux", runtime.GOOS == "darwin":
```
Thykof, line 32: this function was previously `getConfigDir` declined in files `dirs_unix.go` and `dirs_windows.go`. I understand that folder `pkg` wasn't the right place for this function, but why are you assembling this function with `case runtime.GOOS`?

int/configuration/ca_darwin_test.go
Thykof, line 1: here you are writing tests for darwin, in another file for windows, but the test CI only executes on Ubuntu. If we want those tests to be run in the CI, we should change de test CI to add a matrix on OS for `runs-on`.

int/configuration/configuration.go
gregLibert: because it easier to read that to navigate between 3 files as everything enter in one screen page

int/configuration/ca_darwin_test.go
gregLibert: indeed. I think its important to insure that the program is working on all OS. But this is not the subject of this PR I guess. You should create an issue on that as your remark is still relevant.

pkg/certificate/authority.go
Thykof: okay

int/configuration/configuration.go
Thykof: true

int/configuration/ca_darwin_test.go
Thykof: yes #994

Thykof (approved):

B-Naoufal (approved):

PR #981, opened 2023-07-07
add timeout to handle graceful exit

B-Naoufal:


Thykof:
I tested on mac os, the Massa Station app stays open but blocked until the end of the timeout, which is not the expected behavior. In addition, imagine we have so many websites that we must wait 2 minutes... impossible to see the page.

Should we find a way to create a cancelable timeout, and cancel this timeout on app quit?

int/api/server.go
```diff
@@ -91,20 +92,23 @@ func NewServer(flags StartServerFlags) *Server {
...
 
-	return &Server{
-		api:      server,
+	server := &Server{
+		api:      apiServer,
 		localAPI: localAPI,
 		shutdown: make(chan struct{}),
 	}
+
+	return server
```
Thykof, line 111: this change is unecessary, you can directly return the Server object.

and so the renaming `apiServer := restapi.NewServer(localAPI)` is unnecessary too.

Thykof (changes requested):

Thykof (approved):

PR #982, opened 2023-07-07
create websites cache dir on install on windows

Thykof:
This pull request fix a bug on windows, on any browser, navigating to http://flappy.massa/ failed because the websitesCache directory wasn't created and golang code can't do it because it's read only, so this fix creates this directory during installation.

This bug raise to other questions:
- GetConfigDir on windows returns C:\Program Files (x86)\MassaStation whereas it could return C:\Users\X\AppData\Local\MassaStation
- the error wasn't logged into the log file because it's a panic serving of goswagger.

### DIFF
--- a/installer/windows/build_installer.py
+++ b/installer/windows/build_installer.py
@@ -214,6 +214,13 @@ def create_wxs_file():
                             </CreateFolder>
                         </Component>
                     </Directory>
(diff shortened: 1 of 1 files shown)


thomas-senechal (approved):

B-Naoufal (approved):
well done

PR #989, opened 2023-07-10
Update README.md

Thykof:


Thykof:
done here https://github.com/massalabs/station/blob/main/CONTRIBUTING.md?plain=1#L114

PR #991, opened 2023-07-10
990 clean sni

gregLibert:


pkg/certificate/pem.go
```diff
@@ -0,0 +1,119 @@
...
+	case X509CRL:
+		return "X509 CRL"
+	case PrivateKey:
+		return "PRIVATE KEY"
+	default:
+		return ""
+	}
+}
+
+// PemTypeString retrieves an enum value from the enum constants string name.
```
ASAPSegfault, line 77:
```suggestion
// NewPemType retrieves an enum value from the enum constants string name.
```

ASAPSegfault (changes requested):

pkg/certificate/certificate.go
```diff
@@ -1,206 +1,93 @@
...
-	}
-
-	return cert, privateKey, nil
-}
-
-// Get the mkcert CA root path depending on the OS.
-func mkcertCARootPath() (string, error) {
-	if env := os.Getenv("CAROOT"); env != "" {
-		return env, nil
+		return nil, fmt.Errorf("%w at %s: %w", ErrFailedToReadFile, path, err)
```
thomas-senechal, line 48: Looks like you can't have multiple occurences of `%w` in a `fmt.Errorf` and that's why tests are failing

pkg/config/check.go
```diff
@@ -31,3 +40,67 @@ func checkCertificate() error {
...
+	err := store.Add(cert)
+	if err != nil {
+		return fmt.Errorf("failed to add the CA to the operating system: %w", err)
+	}
+
+	return nil
+}
+
+// Get the mkcert CA root path depending on the OS.
+func mkcertCARootPath() (string, error) {
```
peterjah, line 77: I think this func is defined twice (here and in sni.go) maybe its posible to factorize ?

int/sni/sni.go
```diff
@@ -0,0 +1,171 @@
...
+		},
+		NotBefore: time.Now(),
+		NotAfter:  time.Now().AddDate(0, 0, 1),
+		DNSNames:  []string{serverName},
+	}
+
+	return template, nil
+}
+
+// generateSignedCertificate creates a certificate and then signs it using the provided Certificate Authority (CA).
```
Thykof, line 75: it is not explicit how you provide the CA to this function. The CA is read from `mkcertCARootPath`.

thomas-senechal (changes requested):

Thykof (changes requested):

pkg/certificate/certificate.go
gregLibert: > Looks like you can't have multiple occurences of `%w` in a `fmt.Errorf` and that's why tests are failing

This functionality was added in go 1.20.

Why do i need both ? To let errors.As be usable.

pkg/certificate/pem.go
```diff
@@ -0,0 +1,119 @@
...
+		return "X509 CRL"
+	case PrivateKey:
+		return "PRIVATE KEY"
+	default:
+		return ""
+	}
+}
+
+// NewPemType retrieves an enum value from the enum constants string name.
+// Throws an error if the param is not part of the enum.
```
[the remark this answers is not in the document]
gregLibert: I don't understand ...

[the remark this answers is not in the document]
Thykof: ho sorry, it's the wording of the comment: the function doesn't throw but return an error

pkg/config/check.go
gregLibert: It is. And check config shouldn't be a package. In my initial PR everything was done. But it was to big. I propose to live it like tha and then to clean the config/pkg.

int/sni/sni.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	return &tls.Certificate{
+		Certificate: [][]byte{certBytes},
+		PrivateKey:  privateKey,
+	}, nil
+}
+
+// Get the mkcert CA root path depending on the OS.
+func mkcertCARootPath() (string, error) {
```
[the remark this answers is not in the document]
gregLibert: I don't think it's a good idea because check config shouldn't be a package. In my initial PR everything was done. But it was to big. I propose to leave it like that and then to clean the config/pkg.

pkg/certificate/pem.go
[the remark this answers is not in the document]
gregLibert: yes, good catch.

int/sni/sni.go
gregLibert: it's done just after: // It uses root certificate and private key from the default location.
I don't want to mention the mkcertCARootPath because it will be cleaned in a future PR.

pkg/certificate/pem.go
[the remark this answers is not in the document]
gregLibert: Should be fixed in 0d7720a6105c701834753e4dffedb030609fa1d1

pkg/certificate/pem.go
```diff
@@ -0,0 +1,119 @@
...
+func NewPemType(str string) (PemType, error) {
+	switch str {
+	case "CERTIFICATE":
+		return Certificate, nil
+	case "PRIVATE KEY":
+		return PrivateKey, nil
+	default:
```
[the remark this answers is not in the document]
gregLibert: Should be fixed in 0d7720a6105c701834753e4dffedb030609fa1d1

Thykof (approved):
Nice!

thomas-senechal (commented):
Only a last issue related to the linter

Also, when will we update Go to 1.20 ? Do we have an issue for it already ?

peterjah (approved):

ASAPSegfault (approved):

B-Naoufal (approved):

pkg/config/check.go
Thykof: we will have to remove this function during merge conflict resolution (I did it locally for a dry run)

PR #1005, opened 2023-07-13
Add Store Loading State

pivilartisant:
https://github.com/massalabs/station/assets/90157528/f0b6c394-5336-436d-ac9d-fb4af5a21de0

### DIFF
--- a/web/massastation/src/pages/Store/PluginCardLoading.tsx
+++ b/web/massastation/src/pages/Store/PluginCardLoading.tsx
@@ -1,0 +1,20 @@
+import { FiDownload } from 'react-icons/fi';
+import { Plugin } from '@massalabs/react-ui-kit';
+export function PluginCardLoading() {
+  const argsStore = {
+    preIcon: <img src="" alt="plugin-logo" />,
+    topAction: <FiDownload />,
+    title: '',
+    subtitle: '',
+    subtitleIcon: null,
+    content: '',
+  };
+
+  return (
+    <>
+      <div className="w-fit h-fit blur-sm animate-pulse">
+        <Plugin {...argsStore} />
+      </div>
+    </>
+  );
+}
--- a/web/massastation/src/pages/Store/StoreSection/StorePlugin.tsx
+++ b/web/massastation/src/pages/Store/StoreSection/StorePlugin.tsx
@@ -4,15 +4,12 @@ import { usePost } from '@/custom/api';
 import { Certificate, Plugin } from '@massalabs/react-ui-kit';
 import { massalabsNomination } from '@/const';
 import { FiDownload } from 'react-icons/fi';
+import { PluginCardLoading } from '../PluginCardLoading';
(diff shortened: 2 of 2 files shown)


web/massastation/src/pages/Store/PluginCardLoading.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Plugin } from '@massalabs/react-ui-kit';
+export function PluginCardLoading() {
+  const argsStore = {
+    preIcon: <img src="" alt="plugin-logo" />,
+    topAction: <FiDownload />,
+    title: 'Lorem Ipsum Plugin',
+    subtitle: 'Lorem Author',
+    subtitleIcon: null,
+    content: 'Lorem Ipsum Doloris Sit Amet',
+  };
```
Pierregvx, line 11: Do we need this ? it seem to me that we didn't use this in the other loader, tell me if I am wrong

pivilartisant: That's a good point I guess it can be removed

Thykof: now it's not the same result as in you video, is it intended?

pivilartisant: indeed it's a different results : 
[Screencast from 17-07-2023 11:04:17.webm](https://github.com/massalabs/station/assets/90157528/0ddd3e62-611a-4914-8969-ae33166c0bc8)

I would say it a product vision choice @SlnPons  between this one and the original one wdyt ?

SlnPons: @pivilartisant I don't see the 2 options here. What is the original vs. "this one". 
Knowing that I haven't personally prioritise this task... I need more details to validate.

pivilartisant: > @pivilartisant I don't see the 2 options here. What is the original vs. "this one". Knowing that I haven't personally prioritise this task... I need more details to validate.

The original one is in the pr description, the second one is linked in my last comment

SlnPons: I'm sorry @pivilartisant I don't see the difference :/ 
From a product perspective, the most important is to be consistent. So the loading state MUST be the same across all our products. I do believe that the "original" loading state is already used on other pages, if that's the case, we need to use it everywhere.
The loading states on Massa Station have been implemented without my validation so i'm not across the logic here... I can't help more that by saying: it must be the same across all massa station app.

pivilartisant: Hello @SlnPons I've checked with @LucasF243, He green lights we for the second option which is also standard across massa station & wallet :)

Thykof (approved):

PR #1022, opened 2023-07-18
Update sendoperation pkg testnet 23

Thykof:


Thykof:
I closed because work is done here https://github.com/massalabs/station/pull/948

PR #1028, opened 2023-07-19
remove useless flag

B-Naoufal:


thomas-senechal (approved):

thomas-senechal (changes requested):
I approved by mistake

.github/workflows/api.yml
```diff
@@ -3,9 +3,6 @@ name: API
 on:
   push:
 
-env:
-  NODE_SERVER: BUILDNET
-
```
thomas-senechal, line 8: I think we should keep this and use the config file to set the `NODE_SERVER` we want to us.

thomas-senechal (changes requested):

.github/workflows/api.yml
B-Naoufal: @thomas-senechal  IMO that's why we added i've added the yaml file where we set the default NODE_SERVER and DNS to use , for your use case, you will need to just update the config_network.yml file

B-Naoufal: @thomas-senechal also NODE_SERVER  was used in `run: xvfb-run ./massastation --node-server $NODE_SERVER`but the point of this PR is to not use  --node-server  flag anymore ...

thomas-senechal: I'm talking about using the config file in the CI, to allow us to switch network easily if needed

B-Naoufal: can you explain more please ? the config_network.yml file is already used otherwise the ci won't work , everytime massa  station starts , it init the NODE_SERVER with the default value!

README.md
Thykof, line 1: I think we should add a sentence in the contribution to tell that if developer wants to change the network and/or the DNS then It's possible with the yml file.

Thykof (changes requested):

.github/workflows/api.yml
Thykof: will be covered in https://github.com/massalabs/station/issues/1079

Thykof (approved):

thomas-senechal (approved):

PR #1037, opened 2023-07-21
1036 webonchain on firefox windows no favicon in website list

Thykof:


int/api/websites/all.go
```diff
@@ -155,7 +155,7 @@ func DNSRecordFavicon(name, websiteStorerAddress string, client *node.Client) st
 		return ""
 	}
 
-	return "https://" + name + ".massa/" + faviconIcon
+	return "http://" + name + ".massa/" + faviconIcon
```
gregLibert, line 158: Why do you need to return the protocol?

Thykof: good question, maybe we should take the current protocol on the frontend side, I'll check

Thykof: indeed! https://github.com/massalabs/station/pull/1037/commits/b7ff9226eb997bd825618536bff3ed48b0a9c5f5

mazmassa (approved):

PR #1050, opened 2023-07-25
961 add ca to firefox on windows

thomas-senechal:


installer/windows/build_installer.py
```diff
@@ -73,29 +75,15 @@ def build_massastation():
...
-
-    # The previous `fyne package` command generates MassaStation.exe binary in the src directory.
-    # That's why we need to move it in the current directory and rename it to $MASSASTATION_BINARY.
-    os.rename(
-        os.path.join("..", "cmd", "massastation", "MassaStation.exe"),
-        os.path.join(MASSASTATION_BINARY),
-    )
+    subprocess.run([
+        "go",
+        "build",
```
Thykof, line 80: why are you changing fyne package by go build?

pkg/nss/runner_windows.go
```diff
@@ -5,19 +5,45 @@ package nss
...
 func (r *CertUtilRunner) Run(args ...string) error {
-	return fmt.Errorf("not implemented")
+	cmd := exec.Command(r.binaryPath, args...) // #nosec G204
+
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("%s: %w", string(out), err)
+	}
+
+	return nil
```
Thykof, line 48: duplicated code with https://github.com/massalabs/station/blob/e88e04f40afee7af81b101a384d8b43064c5ae7c/pkg/nss/runner_unix.go#L28

pkg/nss/manager_windows.go
```diff
@@ -4,17 +4,60 @@ import (
...
 
+// getFirefoxProfilePath returns the path to the Firefox default profile directory.
+// It will search for the first directory that ends with ".default-release" which is the directory that contains the NSS database.
+func getFirefoxProfilePath() (string, error) {
+	appData := os.Getenv("APPDATA")
+	if appData == "" {
+		return "", fmt.Errorf("failed to get APPDATA environment variable")
+	}
+
+	firefoxProfilesDir := filepath.Join(appData, "Mozilla\\Firefox\\Profiles")
```
Thykof, line 18:
```suggestion
	firefoxProfilesDir := filepath.Join(appData, "Mozilla", "Firefox", "Profiles")
```

pkg/nss/manager_windows.go
```diff
@@ -4,17 +4,60 @@ import (
...
 		return nil, fmt.Errorf("failed to get APPDATA environment variable")
 	}
 
-	nssDBPaths := []string{
-		filepath.Join(appData, "Mozilla\\Firefox\\Profiles"),
+	nssDBPaths := []string{}
+
+	firefoxDBPath, err := getFirefoxProfilePath()
+	if err == nil {
+		nssDBPaths = append(nssDBPaths, firefoxDBPath)
```
Thykof, line 60: this will fail silently, is it intended? if yes, should we add a comment to explain why?

pkg/nss/certutilservice.go
```diff
@@ -11,3 +11,31 @@ package nss
...
+
+// NewCertUtilService returns a new CertUtilService.
+func NewCertUtilService(runner Runner) (*CertUtilService, error) {
+	return &CertUtilService{runner: runner}, nil
+}
+
+// AddCA adds a certificate to the NSS database CA list.
+func (s *CertUtilService) AddCA(dbPath string, certificateName string, certPath string) error {
+	//nolint:wrapcheck
+	return s.runner.Run("-A", "-d", dbPath, "-t", "C,,", "-n", certificateName, "-i", certPath)
```
Thykof, line 28: ho it's the same parameters for certutils and mar-tools!

Thykof (changes requested):

installer/windows/build_installer.py
thomas-senechal: To match what we do in the CI for Windows

pkg/nss/runner_windows.go
thomas-senechal: Indeed, but since those functions are for 2 different structures, I don't see how we can fix this 
Do you have an idea ?

pkg/nss/manager_windows.go
thomas-senechal: It's just that there is no need to add a path related to Firefox if Firefox isn't installed
And since this code runs at each start up of MassaStation, it will check at each start if Firefox is installed, and if so, it will automagically fix the missing CA root for it

pkg/nss/certutilservice.go
thomas-senechal: In fact, `mar-tools` embeds `certutils`, so, we are using the same binary (at lest, they work the same way)

pkg/nss/manager_windows.go
Thykof: then add a comment "add Firefox NSS path if exists" for example

pkg/nss/runner_windows.go
Thykof: maybe by moving the struct declaration into runner.go, move also that function, keep NewCertUtilRunner in OS specific go file.

Thykof: @gregLibert does this remark block from merging? because otherwise LGTM

Thykof (approved):

pkg/nss/manager_windows.go
```diff
@@ -4,17 +4,62 @@ import (
...
-	nssDBPaths := []string{
-		filepath.Join(appData, "Mozilla\\Firefox\\Profiles"),
+	nssDBPaths := []string{}
+
+	firefoxDBPath, err := getFirefoxProfilePath()
+	if err == nil {
+		// If we found the Firefox profile directory, we add it to the list of NSS databases.
+		// If we didn't find it, we just ignore it and continue because it means that Firefox is not installed.
+		nssDBPaths = append(nssDBPaths, firefoxDBPath)
 	}
```
gregLibert, line 63: Why aren't you simply doing:
```suggestion
	nssDBPaths := []string{
		filepath.Join(appData, "Mozilla", "Firefox", "Profiles", "*"),
	}
```

pkg/nss/runner_windows.go
```diff
@@ -5,19 +5,45 @@ package nss
...
 import (
 	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
 )
 
 // CertUtilRunner encapsulates certutil commands.
-type CertUtilRunner struct{}
+type CertUtilRunner struct {
```
gregLibert, line 14: should be in runner.go is the code is the same for unix

pkg/nss/runner_windows.go
```diff
@@ -5,19 +5,45 @@ package nss
...
+	binaryPath string
+}
 
 // NewCertUtilRunner returns a new CertUtilRunner.
 // It returns an error if the certutil binary is not found.
 func NewCertUtilRunner() (*CertUtilRunner, error) {
-	return nil, fmt.Errorf("not implemented")
+	ex, err := os.Executable()
```
gregLibert, line 21:
```suggestion
	pathExecutable, err := os.Executable()
```

pkg/nss/runner_windows.go
```diff
@@ -5,19 +5,45 @@ package nss
...
+	if err != nil {
+		return nil, fmt.Errorf("getting executable path: %w", err)
+	}
+
+	dir := filepath.Dir(ex)
```
gregLibert, line 26:
```suggestion
	pathDirectory := filepath.Dir(ex)
```

pkg/nss/runner_windows.go
```diff
@@ -5,19 +5,45 @@ package nss
...
+	bin := filepath.Join(dir, "mar-tools", "certutil.exe")
```
gregLibert, line 27:
```suggestion
	pathCertutil := filepath.Join(dir, "mar-tools", "certutil.exe")
```

pkg/nss/runner_windows.go
```diff
@@ -5,19 +5,45 @@ package nss
...
+	}
+
+	return &CertUtilRunner{binaryPath: bin}, nil
+
 }
 
 // Run runs the certutil command with the given arguments.
-// It returns the combined output of stdout and stderr.
+// It returns the combined output of stdout and stderr in the error, if any.
 func (r *CertUtilRunner) Run(args ...string) error {
```
gregLibert, line 40: Should be in runner.go as the code is the same as on unix

gregLibert (changes requested):

pkg/nss/manager_windows.go
thomas-senechal: Because certutil returns an error when using `*`. I don't know if it's because of the other profile in the directory, or just because it does not find the "*" directory and so, fails.

image

thomas-senechal: It looks like it's because certutil searches for the `*` directory. 
I tried with a directory name that does not exists and got the same error

gregLibert: Why is it working for linux ? expander [here](https://github.com/massalabs/station/blob/f2e244862f63372c0f942440b7ce8371ed3ce768/pkg/nss/manager.go#L116) should accept a *

gregLibert (approved):

PR #1051, opened 2023-07-25
fix plugin update already registered error

peterjah:


Pierregvx (approved):

pkg/plugin/manager.go
```diff
@@ -117,9 +117,6 @@ func (m *Manager) PluginByAlias(alias string) (*Plugin, error) {
 
 // Plugin returns a plugin from the manager.
 func (m *Manager) Plugin(correlationID string) (*Plugin, error) {
-	m.mutex.RLock()
```
gregLibert, line 120: I think that if you remove the lock, you have a race condition error when:
- I call the Plugin function
- at the same time the plugin is delete

gregLibert (commented):
Sounds good. The race condition is not really important because it's probability is very low. I think it's safe to move this subject to the test actions to plan in the future.

pkg/plugin/manager.go
peterjah: nice catch! i will put back the lock

PR #1057, opened 2023-07-27
generate new CA

Thykof:


pkg/certificate/authority.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// writeCA writes the private key and certificate to the specified path.
+func writeCA(path string, privateKey crypto.PrivateKey, cert []byte) error {
+	privDER, err := x509.MarshalPKCS8PrivateKey(privateKey)
+	if err != nil {
+		return fmt.Errorf("failed to marshal private key: %w", err)
+	}
+
+	permission := 777
```
Thykof, line 123: on macos, the destination directory has a umask of 022. How can I improve this?

Thykof: solve, os.FileMode expect octal, so 0o644

installer/deb/scripts/postinst
```diff
@@ -73,36 +70,12 @@ set_local_dns() {
...
-    fi
-
-    ./mkcert -install || fatal "mkcert installation failed."
-    mkdir -m 777 -p $MASSASTATION_CERT_DIR || fatal "certs directory creation failed."
-    ./mkcert -cert-file $MASSASTATION_CERT_DIR/cert.pem -key-file $MASSASTATION_CERT_DIR/cert-key.pem station.massa || fatal "mkcert certificate generation failed."
-    chmod 644 $MASSASTATION_CERT_DIR/cert.pem $MASSASTATION_CERT_DIR/cert-key.pem || fatal "certs permission change failed."
-
-    rm mkcert || fatal "mkcert removal failed."
-}
-
```
thomas-senechal, line 95: Didn't we said that we would keep this until we are 100% sure everything is working as expected ? 🤔

Thykof: we moved to plan B ^^

pkg/certificate/authority.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+const (
+	CertificateAuthorityFileName    = "rootCA.pem"
+	CertificateAuthorityKeyFileName = "rootCA-key.pem"
+)
+
+// randomSerialNumber generates a random serial number for the certificate.
+func randomSerialNumber() (*big.Int, error) {
+	n := 128
```
gregLibert, line 27:
```suggestion
bitSize := 128
```

pkg/certificate/authority.go
```diff
@@ -0,0 +1,173 @@
...
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), uint(n))
```
gregLibert, line 28: You should add a comment such as :
```
// Calculate the limit as 2^bitSize
```

gregLibert (changes requested):

int/config/check.go
```diff
@@ -55,12 +72,21 @@ func caNonBlockingError(context string, err error) error {
...
 // checkCertificate checks the certificate configuration.
-func checkCertificate(certPath string) error {
+func checkCertificate(certPath string, cleanCA bool) error {
 	certCA, err := certificate.LoadCertificate(certPath)
 	if err != nil {
 		return fmt.Errorf("failed to load the CA: %w", err)
 	}
 
+	if cleanCA {
+		err = store.Delete(certCA)
```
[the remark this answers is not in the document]
Thykof: for NSS database, if we generate a brand new CA, it's probable that a CA with the same name (massaStation) already exists in NSS database. So first we delete it, then add it. However, I'm not sure for platform database but I did the same logic.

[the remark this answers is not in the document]
gregLibert: ok, let's keep it as a security. Can you add a comment explaining this ? Because I guess it's only temporary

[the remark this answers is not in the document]
Thykof: comment added.

gregLibert (approved):

pkg/certificate/authority.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		return err
+	}
+
+	publicKey, err := derivePublicKey(privateKey)
+	if err != nil {
+		return fmt.Errorf("failed to derive public key: %w", err)
+	}
+
+	template, err := createCATemplate(serialNumber, publicKey, "MassaLabs")
```
thomas-senechal, line 159: I would make that "MassaLabs" as a const variable

thomas-senechal (approved):
Good job! Just a non blocking comment

I tested on Windows and it's working

Thykof:
it works on ubuntu

Thykof:
it works on macos

Pierregvx:
Worked well on my Ubuntu vm & host and on my windows 11 dual boot

PR #1082, opened 2023-08-08
add a decrypter for string sent to the wallet signer

B-Naoufal:


pkg/node/sendoperation/serializeaddress/serializeaddress.go
```diff
@@ -29,3 +31,25 @@ func SerializeAddress(addr string) ([]byte, error) {
 
 	return append(result, addressBytes...), nil
 }
+
+func DeserializeAddress(versionedAddress []byte) (string, error) {
```
Thykof, line 35: add unit test for SerializeAddress and DeserializeAddress

pkg/node/sendoperation/sendoperation.go
Thykof, line 1: move DecodeCallSCMessage and GetCallSCFromSign to callsc.go

pkg/node/sendoperation/callsc/callsc.go
```diff
@@ -74,6 +75,10 @@ func (c *CallSC) Message() []byte {
 	nbBytes = binary.PutUvarint(buf, c.coins)
 	msg = append(msg, buf[:nbBytes]...)
 
+	// target address length
+	nbBytes = binary.PutUvarint(buf, uint64(len(c.address)))
+	msg = append(msg, buf[:nbBytes]...)
+
```
Thykof, line 81: why add this? does the node can handle it?

Thykof (changes requested):

pkg/node/sendoperation/callsc/callsc.go
```diff
@@ -89,3 +105,84 @@ func (c *CallSC) Message() []byte {
 
 	return msg
 }
+
+//nolint:funlen, cyclop
+func DecodeCallSCMessage(data []byte) (*FromSign, error) {
```
Thykof, line 110: rename that function DecodeMessage

Thykof (changes requested):

pkg/node/sendoperation/callsc/callsc.go
```diff
@@ -47,11 +54,16 @@ func New(address string, function string, parameters []byte, gasLimit uint64, co
 }
 
 func (c *CallSC) Content() interface{} {
+	addressString, err := utils.DeserializeAddress(c.address)
+	if err != nil {
+		return nil
```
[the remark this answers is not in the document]
B-Naoufal: actually we don't need it, since we deserialize the same address that was successfully serialized in creation of the operation in. 
`func New(address string, function string, parameters []byte, gasLimit uint64, coins uint64,
) (*CallSC, error) {
	versionedAddress, err := utils.SerializeAddress(address)` so no need to check this error , it's an unreachable code.

pkg/node/sendoperation/callsc/callsc.go
B-Naoufal: i don't know if the node ever decode this message, but there is no way to decode the address without adding the target address length to the message , and you can check the unit tests that the serialization and deserialization is working with this added  target address length .

pkg/node/sendoperation/callsc/callsc.go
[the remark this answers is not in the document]
Thykof: no, you can create an operation in another software that Massa Station (massa-web3 for example). We should not do the hypothesis of unreachable code. When you call `c.Content()` we don't want to do this:

```go
content := c.Content()
if content == nil {
  panic("...")
}
```

but rather:

```go
content, err := c.Content()
if err != nil {
  panic("...")
}
```

pkg/node/sendoperation/callsc/callsc.go
Thykof: The api e2e tests are failing with "Failed target_addr deserialization".

pkg/node/sendoperation/callsc/callsc.go
```diff
@@ -89,3 +105,84 @@ func (c *CallSC) Message() []byte {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read address length: %w", err)
+	}
+
+	address := make([]byte, addressLength)
+
+	_, err = buf.Read(address)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read address: %w", err)
+	}
```
Thykof, line 149:
```suggestion
	assert.NoError(t, err)
```

Thykof (changes requested):

pkg/node/sendoperation/callsc/callsc.go
B-Naoufal: why assert , this is [callsc.go], why we should add the assert ?

Thykof: you're right, I would like to show an example of the assert package, related to https://github.com/massalabs/station/pull/1082#discussion_r1288536795 but I should have done this in a unit test file

pkg/node/sendoperation/serializeaddress/serializeaddress_test.go
```diff
@@ -64,3 +64,38 @@ func bytesEqual(byteA, byteB []byte) bool {
...
+			expectedAddress: "AU1MPDRXuR22mwYDFCeZUDgYjcTAF1co6xujx2X6ugoHeYeGY3B5",
+			expectError:     false,
+		},
+	}
+
+	for _, testCase := range testCases {
+		address, err := DeserializeAddress(testCase.versionedAddress)
+		if testCase.expectError {
+			if err == nil {
+				t.Errorf("Expected an error but did not receive one for versioned address %v", testCase.versionedAddress)
```
Thykof, line 88:
```suggestion
			assert.Error(t, err, fmt.Sprintf("Expected an error but did not receive one for address %s", testCase.addr))

```

pkg/node/sendoperation/callsc/callsc.go
Thykof: > there is no way to decode the address without adding the target address length to the message 

there is one in rust, on massa core side, so we must find this same way in golang

pkg/node/sendoperation/serializeaddress/serializeaddress.go
```diff
@@ -29,3 +31,30 @@ func SerializeAddress(addr string) ([]byte, error) {
...
+	const minAddressLength = 2
```
peterjah, line 36: from the discord conversation its seems min length is  34 . isn't it ?

Thykof: yes

pkg/node/sendoperation/callsc/callsc.go
```diff
@@ -8,7 +8,10 @@ import (
 	utils "github.com/massalabs/station/pkg/node/sendoperation/serializeaddress"
 )
 
-const CallSCOpID = uint64(4)
+const (
+	CallSCOpID                      = uint64(4)
+	addressWithoutPrefixVersionSize = 32
```
Thykof, line 13: maybe name it publicKeyHashSize? because it's address without type and version prefix actually.

pkg/node/sendoperation/callsc/callsc.go
```diff
@@ -135,25 +134,37 @@ func DecodeMessage(data []byte) (*MessageContent, error) {
 
 	callSCContent.Coins = coins
 
-	// Read target address length
-	addressLength, err := binary.ReadUvarint(buf)
+	// Read address prefix
+	addressPrefix, err := binary.ReadUvarint(buf)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read address prefix: %w", err)
+	}
```
Thykof, line 141: it's the address type

Thykof (changes requested):

pkg/node/sendoperation/callsc/callsc.go
Thykof, line 1: in Message function, maybe we should encode the address type and the address version separately with `PutUvarint`?

Thykof (approved):

peterjah (approved):

PR #1086, opened 2023-08-10
979 move dirs package to int

Thykof:


cmd/massastation/main.go
```diff
@@ -80,7 +80,7 @@ func main() {
 		server.Stop()
 	})
 	stationGUI.Lifecycle().SetOnStarted(func() {
-		server.Start(networkManager, pluginManager)
+		server.Start(networkManager, pluginManager, configDir)
```
peterjah, line 83: why passing this configDir everywhere ? instead of calling `configuration.Path()` where you need it ?

Thykof: because I can't import int package into pkg packages

thomas-senechal (approved):

api/middleware.go
```diff
@@ -19,10 +23,19 @@ func TopMiddleware(handler http.Handler, cfg config.AppConfig) http.Handler {
 	//nolint:varnamelen
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		logger.Infof("[%s %s]", r.Method, r.URL.Path)
+		if configDir == "" {
+			var err error
+			configDir, err = configuration.CertPath()
```
peterjah, line 28: `configDir = certPath` ? if yes maybe some renaming is needed to clarify

peterjah (approved):

PR #1093, opened 2023-08-11
fix switch network on firefox

Thykof:


maxime-carabina (approved):
Nice fix 😉 
LGTM 🎉

thomas-senechal:
Not working on my MacOS laptop :/

The behavior seems to be really random tho, sometimes, I select another network and it changes well, I refresh the page and it's still the right network, and then, if I change again, it fails :/

Also, I noticed that the Network order in the select is random, I checked what returned the `GET /network` endpoint, and the list is also ordered randomly. Looking at how we handle network changes, this does not seems to be the issue and that it's working fine on other web browsers

thomas-senechal (approved):

PR #1095, opened 2023-08-11
fix nickname not  in account list in web on chain

Pierregvx:


web/massastation/src/components/SidePanels/Upload/Upload.tsx
```diff
@@ -185,6 +185,10 @@ export default function Upload() {
   const { data: accounts = [] } = useResource<AccountObject[]>(
     `${URL.WALLET_BASE_API}/${URL.WALLET_ACCOUNTS}`,
   );
+  const nicknameInAccounts = accounts.find(
+    (account) => account.nickname === nickname,
+  );
+  !nicknameInAccounts && setNickname(accounts[0]?.nickname || '');
```
Thykof, line 191: write a if statement, it's clearer IMO

Thykof (approved):

web/massastation/src/components/SidePanels/Upload/Upload.tsx
```diff
@@ -185,6 +185,10 @@ export default function Upload() {
   const { data: accounts = [] } = useResource<AccountObject[]>(
     `${URL.WALLET_BASE_API}/${URL.WALLET_ACCOUNTS}`,
   );
+  const nicknameInAccounts = accounts.find(
+    (account) => account.nickname === nickname,
+  );
+  if (!nicknameInAccounts) setNickname(accounts[0]?.nickname || '');
```
maxime-carabina, line 191: You can't set an `''` value when the user can't be found. 🤔 

What can we do when `accounts[0]?.nickname` is undefined?

web/massastation/src/components/SidePanels/Upload/Upload.tsx
```diff
@@ -185,6 +185,10 @@ export default function Upload() {
   const { data: accounts = [] } = useResource<AccountObject[]>(
     `${URL.WALLET_BASE_API}/${URL.WALLET_ACCOUNTS}`,
   );
+  const nicknameInAccounts = accounts.find(
```
maxime-carabina, line 188:
```suggestion
  const nicknameExistInAccountsList = accounts.find(
```

More explicit 😉

web/massastation/src/components/SidePanels/Upload/Upload.tsx
Pierregvx: technically this does not happen because There is a `create an account` button, ill see how I can make this better

web/massastation/src/components/SidePanels/Upload/Upload.tsx
Thykof: IMO for me nicknameExistInAccountsList means this variable is a boolean, true if the nickname exists in account list.

web/massastation/src/components/SidePanels/Upload/Upload.tsx
maxime-carabina: > technically this does not happen because There is a `create an account` button, ill see how I can make this better

Sure! But be careful, as this condition could create UI problems in the future.

**EX:**  The user decides to open the left panel, but (for some reason) his name doesn't appear in the list.

Pierregvx: I added a condition in the if to avoid doing this

maxime-carabina: > I added a condition in the if to avoid doing this

Great job! Thanks for that! 🎉 😉

maxime-carabina (approved):

PR #1108, opened 2023-08-18
remove useless server flag

Thykof:


thomas-senechal (approved):

int/api/server.go
```diff
@@ -23,23 +21,14 @@ import (
 )
 
 type StartServerFlags struct {
-	Port              int
-	TLSPort           int
-	TLSCertificate    string
-	TLSCertificateKey string
+	Port    int
+	TLSPort int
+	Version bool
```
thomas-senechal, line 26: I think the `Version` isn't supposed to be there since we moved it to https://github.com/massalabs/station/blob/3046b23272ed586f47446e37d48ae0560ba0cb9a/cmd/massastation/main.go#L19C2-L19C2

thomas-senechal (approved):

PR #1110, opened 2023-08-21
add decoder for roll buy and roll sell

B-Naoufal:


Thykof (approved):
it can be nice to have a curl command to test that

pkg/node/sendoperation/buyrolls/buy_rolls.go
```diff
@@ -47,3 +49,23 @@ func (b *BuyRolls) Message() []byte {
 
 	return append(msg, buf[:nbBytes]...)
 }
+
+func DecodeMessage(data []byte) (*OperationDetails, error) {
+	buf := bytes.NewReader(data)
+
+	// Read operationId
```
thomas-senechal, line 56: For consistency with sell_roll
```suggestion
	// Skip operationId
```

pkg/node/sendoperation/buyrolls/buy_rolls.go
```diff
@@ -47,3 +49,23 @@ func (b *BuyRolls) Message() []byte {
 
 	return append(msg, buf[:nbBytes]...)
 }
+
+func DecodeMessage(data []byte) (*OperationDetails, error) {
```
thomas-senechal, line 53: This function looks exactly the same as the one in `sell_rolls`, shouldn't we move it in the `sendoperation` package to use the it at both places and avoid duplication ?

thomas-senechal (changes requested):

pkg/node/sendoperation/buyrolls/buy_rolls.go
Thykof: if we move it, we should move it into a new file `rolls.go`

PR #1115, opened 2023-08-23
1097 request for immediate operation id retrieval on smart contract calls

thomas-senechal:


pkg/onchain/sc.go
```diff
@@ -22,20 +22,21 @@ type OperationWithEventResponse struct {
...
-listen to these events in MassaStation and return them as a response.
-Hence this function also listen for the first event emitted in an OP and returns it.
-*/
+// CallFunction calls a function of a smart contract on the blockchain. It returns the event generated
+// by the smart contract and an Error.
+// If `async` is true, it returns the operation response and does not wait for the event.
+// Otherwise, it waits for the first event generated by the smart contract and returns it.
+//
+// Note: Listening to the first event generated by the smart contract should be done by the front end,
+// but for now, it is done here to match the previous implementation.
```
Thykof, line 31: I'm not sure about that...

Thykof (changes requested):

int/api/cmd/execute_function.go
```diff
@@ -41,6 +41,8 @@ func (e *executeFunction) Handle(params operations.CmdExecuteFunctionParams) mid
 		params.Body.Name,
 		args,
 		uint64(params.Body.Coins),
+		uint64(*params.Body.Expiry),
```
Thykof, line 44: nice one! this will partially solve https://github.com/massalabs/station/issues/700

pkg/onchain/sc.go
Thykof: yes, but now that we implement parameter async bool, does this comment still make sense?

thomas-senechal: Hum yes you are right, we could remove it

B-Naoufal (approved):

PR #1117, opened 2023-08-24
return expiry and fee from messageDecoder

B-Naoufal:


pkg/node/sendoperation/sendoperation.go
```diff
@@ -177,30 +177,30 @@ func message(expiry uint64, fee uint64, operation Operation) []byte {
 	return msg
 }
 
-func DecodeMessage64(msgB64 string) ([]byte, error) {
+func DecodeMessage64(msgB64 string) ([]byte, uint64, uint64, error) {
```
Thykof, line 180: could you add a function comment godoc here to explain what are the returned values?

pkg/node/sendoperation/sendoperation_test.go
```diff
@@ -43,7 +43,7 @@ func TestSerializeDeserializeCallSCMessage(t *testing.T) {
 		msgB64 := base64.StdEncoding.EncodeToString(msg)
 
 		// Simulate decoding and deserialization
-		decodedMsg, err := DecodeMessage64(msgB64)
+		decodedMsg, _, _, err := DecodeMessage64(msgB64)
```
Thykof, line 46: can we assert the value of fee and expiry?

Thykof (changes requested):

Thykof (approved):

thomas-senechal (approved):

PR #1145, opened 2023-09-04
1144 cannot send empty string as params for a callsc

thomas-senechal:


thomas-senechal:
I tested it manually since we need to update the wallet to fix the issue

You can test it by adding a `replace` statement in the wallet `go.mod` such as:
```
replace github.com/massalabs/station => ${path to station repo}
```

To be able to build the wallet, you will also need to update `node.DatastoreEntry` to `node.FetchDatastoreEntry` at [station-massa-wallet pkg/network/data_store.go#L82](https://github.com/massalabs/station-massa-wallet/blob/main/pkg/network/data_store.go#L82)

PR #1146, opened 2023-09-05
Refactor webonchain

peterjah:
Achtung:
This is breaking change for webonchain. MS will not able to browse already uploaded websites.
We must deploy a new DNS Contract when merging this


int/api/server.go
```diff
@@ -130,9 +130,8 @@ func (server *Server) printNodeVersion(networkManager *config.NetworkManager) {
 	client := node.NewClient(networkManager.Network().NodeURL)
 	status, err := node.Status(client)
 
-	nodeVersion := "unknown"
 	if err == nil {
-		nodeVersion = *status.Version
+		nodeVersion := *status.Version
```
gregLibert, line 134: Do you need this variable ? Not sure it's needed to understand the code.

pkg/node/sendoperation/sendoperation.go
```diff
@@ -24,6 +24,9 @@ const AccountCreationStorageCost = 10_000_000
 
 const StorageCostPerByte = 1_000_000
 
+// To be updated when key creation cost will be charged for actual size.
+const StorageKeyCreationCost = 10 * StorageCostPerByte
```
gregLibert, line 28: Why 10 ?

pkg/node/sendoperation/sendoperation.go
```diff
@@ -15,7 +15,7 @@ import (
 
 const DefaultGasLimit = 700_000_000
 
-const DefaultSlotsDuration = 2
+const DefaultExpiry = 3
```
gregLibert, line 18:
```suggestion
const DefaultExpiryInSlot = 3
```

pkg/onchain/sc.go
```diff
@@ -15,7 +16,7 @@ import (
 
 const maxWaitingTimeInSeconds = 45
 
-const evenHeartbeat = 2
+const evenHeartbeat = 1
```
gregLibert, line 19: What is variable doing ? 1 is not even, it's odd...

pkg/onchain/sc.go
```diff
@@ -103,57 +113,62 @@ func DeploySC(client *node.Client,
...
+) ([]node.Event, error) {
 	counter := 0
 
 	ticker := time.NewTicker(time.Second * evenHeartbeat)
 
 	for ; true; <-ticker.C {
 		counter++
 
-		if counter > maxWaitingTimeInSeconds*evenHeartbeat {
+		if counter > maxWaitingTimeInSeconds/evenHeartbeat {
```
gregLibert, line 138: What is the point of dividing by 1 ?

pkg/onchain/website/uploader.go
```diff
@@ -109,57 +120,47 @@ func upload(
...
 
 		// Chunk data encoding
-		params = append(params, chunks[index]...)
+		params = append(params, chunks[chunkIndex]...)
+
+		// Total cost is the cost for storage bytes plus a fixed cost for key creation
+		uploadCost := sendOperation.StorageKeyCreationCost + sendOperation.StorageCostPerByte*chunkSize
+
+		if chunkIndex == 0 {
+			// if chunkID == 0, we may need to add the cost of the key creation for the NB_CHUNKS key
```
gregLibert, line 140: Why the may ? Isn't it mandatory?

gregLibert (changes requested):

pkg/node/sendoperation/sendoperation.go
peterjah: its the fixed storage cost of key creation in datastore

pkg/onchain/sc.go
peterjah: If we change the polling interval constant (evenHeartbeat) this code will still work

pkg/onchain/website/uploader.go
peterjah: Its bad copy pasta from the UploadMissedChunks! removing the "may".
in the missed chunk upload context, the NB_CHUNKS will already exist.
But the value we provide here is for "maxCoins" so it it already exists, the coins will not be spent

pkg/node/sendoperation/sendoperation.go
gregLibert: Maybe be you should add a comment or even better replace 10 with an explicite constant name

gregLibert (approved):

PR #1152, opened 2023-09-07
CI: no suffix -dev

Thykof:


thomas-senechal (approved):

Thykof:
installation of the .msi doesn't work windows

thomas-senechal:
I think we should remove the suffix here: https://github.com/massalabs/station/blob/main/pkg/store/store.go#L193C4-L193C4
So when we check for station version >= a plugin version, we won't have the `-dev` version

peterjah (approved):

Thykof:
QA: I can install plugins on Massa Station installed from the artifact of the build dev CI

PR #1165, opened 2023-09-13
taskfile: add end to end test

Thykof:


Taskfile.yml
```diff
@@ -27,6 +27,10 @@ tasks:
     cmds:
       - cmd: go test -v -timeout 60s ./...
 
+  test-2e2:
+    cmds:
```
peterjah, line 31: `platforms: [linux, darwin]`
bash will not work on windows :'(

peterjah: not sure about this one ^^

peterjah (approved):

Taskfile.yml
Thykof: in the readme/contributing we recommend to run task command on a git bash terminal

peterjah (approved):

PR #1179, opened 2023-09-19
fix macos plugin update

peterjah:


Thykof (approved):
It's works, let's wait for another solution if we find a better one before merge

Thykof:
we have #1180 that it maybe a better solution

PR #1189, opened 2023-09-20
fix plugin crash handeling

peterjah:
# backend
- [x] fix is done and tested on the 3 OS

# frontend
- [ ] need to add warning on module page: https://www.figma.com/file/LSaOibVDFb1reF6FePVn5l/Massastation?type=design&node-id=1026-11415&mode=design&t=0yHui61t88i9aoXU-0
- [ ] need to add a toaster on welcome page is wallet if crashed: https://www.figma.com/file/LSaOibVDFb1reF6FePVn5l/Massastation?type=design&node-id=6375-39357&mode=design&t=pUK9E6TXwS5Y8Vq8-0


pkg/plugin/manager.go
```diff
@@ -89,18 +89,11 @@ func (m *Manager) SetAlias(alias string, correlationID string) error {
...
 
-func (m *Manager) RemoveAlias(alias string) error {
+func (m *Manager) RemoveAlias(alias string) {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
 
-	_, exist := m.authorNameToID[alias]
-	if !exist {
-		return fmt.Errorf("while removing alias for %s: no plugin matching the given alias", alias)
-	}
```
Thykof, line 99: I'm not sure about that, but the idea is to not block the rest of the process

web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
```diff
@@ -97,32 +103,73 @@ export function StationPlugin({
...
+        subtitleIcon: massalabsNomination.includes(author) ? (
+          <Certificate />
+        ) : null,
+        version: `v${version}`,
+        content: [
+          <Tooltip
+            className="mas-tooltip"
+            content="The plugin is not working, please uninstall it and install it again."
+            icon={<FiAlertCircle className="text-s-error" />}
+          />,
```
Thykof, line 160: @pivilartisant this is not what design requires, could you have a look? I think we need to modify the uikit

pkg/plugin/manager.go
peterjah: i tried and it seems delete does not error if the key we are deleting does not exist! so you are right this check is useless

web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
pivilartisant: Screenshot 2023-09-21 at 11 14 50

It's looking good to me, what are the modifications you would like to do ?

B-Naoufal (approved):

web/massastation/src/pages/Store/StationSection/StationPlugin.tsx
peterjah: text has been modified

PR #1201, opened 2023-09-27
sign the .msi in the CI

Thykof:
## Delivery Quality Checklist

- [x] **Breaking Changes in API:**

- [x] **Changelog:**

- [x] **PR Dependency:**

- [ ] **Documentation:**


.github/workflows/release.yml
```diff
@@ -28,6 +28,12 @@ jobs:
     uses: ./.github/workflows/build.yml
     with:
       tag_name: ${{ github.event.inputs.release-version }}
+    secrets:
+      AZURE_KEY_VAULT_URI: ${{ secrets.AZURE_KEY_VAULT_URI }}
+      AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+      AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
+      AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
+      AZURE_CERT_NAME: ${{ secrets.AZURE_CERT_NAME }}
```
Thykof, line 36:
```suggestion
```

installer/windows/build_installer.py
```diff
@@ -463,13 +463,14 @@ def build_installer():
     except err:
         print(err)
         sys.exit(1)
+    print(f"Installer built successfully: massastation_{VERSION}_amd64.msi")
```
gregLibert, line 466: debug ?

.github/workflows/build.yml
```diff
@@ -125,27 +126,52 @@ jobs:
...
+      - name: Fetch all git tags
+        run: git fetch --prune --unshallow --tags
+
+      - name: Set Version
+        if: ${{ env.VERSION == '' }}
+        shell: bash
+        run: |
+          export VERSION=$(git describe --tags --abbrev=0 | sed 's/^v//')
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+
```
gregLibert, line 146: Why is it here ? To resolve conflict ?

gregLibert (changes requested):

.github/workflows/build.yml
thomas-senechal: If it's a release, the version is from the input, but otherwise, we use the latest tag as the version

installer/windows/build_installer.py
thomas-senechal: Looks like so
But I think it's not a bad addition and so we can keep it

.github/workflows/build.yml
gregLibert: Why do we have all that here :
```
      - name: Fetch all git tags
        run: git fetch --prune --unshallow --tags

      - name: Set Version
        if: ${{ env.VERSION == '' }}
        shell: bash
        run: |
          export VERSION=$(git describe --tags --abbrev=0 | sed 's/^v//')
          echo "VERSION=$VERSION" >> $GITHUB_ENV
```
is it linked somehow with the need of signing a binary file ?

thomas-senechal: Yes, because we need to know the version for the file name to sign: https://github.com/massalabs/station/pull/1201/files/41ba2e8f15b86e07b8aaed3867c5ca655c422342#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R168

Thykof (commented):
I approve.

I tested to install the .msi, no warning, I installed the .zip plugin build with this action, works perfectly

gregLibert (approved):

PR #1203, opened 2023-09-29
Massa Station Linux HTTPS workaround

eguoli:
This is a common problem for Linux Massa Station users - https protocol is not available by default without adding CA to the browser.

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -80,7 +80,7 @@ Congratulations! You have successfully installed **Massa Station** on your compu
 5. Proceed to GUI installation starting at step 2. 
 
 
-### Linux HTTPS workaround
+### Manually install the HTTPS certificate
 1. Open your browser "Manage Certificates" settings (e.g. chrome://settings/certificates or brave://settings/certificates) and switch to the "Authorities" tab.
 2. Click "Import", locate the Massastation certificate at **/etc/massastation/certs/rootCA.pem** and import it to the browser.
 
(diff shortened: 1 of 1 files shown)


mazmassa (approved):

INSTALLATION.md
```diff
@@ -80,6 +80,11 @@ Congratulations! You have successfully installed **Massa Station** on your compu
 5. Proceed to GUI installation starting at step 2. 
 
 
+### Linux HTTPS workaround
```
Thykof, line 83: I would suggest to change the title:
```suggestion
### Manually install the HTTPS certificate
```

INSTALLATION.md
```diff
@@ -80,6 +80,11 @@ Congratulations! You have successfully installed **Massa Station** on your compu
...
+1. Open your browser "Manage Certificates" settings (e.g. chrome://settings/certificates or brave://settings/certificates) and switch to the "Authorities" tab.
```
Thykof, line 84: I think we could add a sentence explain why user could need this manual step.

Thykof (commented):
Thank you for your contribution! Really appreciated! On Linux, with Chrome browser, latest version of Massa Station is supposed to work with HTTPS without any manual certificate installation. However, this documentation steps are useful for any other browser: explain to user that it requires to manually install the certificate onto the browser.

INSTALLATION.md
eguoli: Good point!

PR #1231, opened 2023-10-12
Windows installer remove terminal windows

thomas-senechal:
## Delivery Quality Checklist

- [x] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

### DIFF
--- a/installer/windows/build_installer.py
+++ b/installer/windows/build_installer.py
@@ -303,111 +303,141 @@ def create_wxs_file():
             Return='ignore'
         />
 
-        <CustomAction Id="ExtractMartools"
-            Directory="INSTALLDIR"
-            ExeCommand="powershell.exe -Command &quot;Expand-Archive '[MassaStationMartools]{MARTOOLS_ZIP}' -d '[INSTALLDIR]'&quot;"
+        <SetProperty Id="ExtractMartools" Value="&quot;powershell.exe&quot; -Command &quot;Expand-Archive '[MassaStationMartools]{MARTOOLS_ZIP}' -d '[INSTALLDIR]'&quot;" Before="ExtractMartools" Sequence="execute" />
+        <CustomAction
+            Id="ExtractMartools"
+            BinaryKey="WixCA"
+            DllEntry="CAQuietExec"
             Execute="deferred"
             Impersonate="no"
             Return="check"
         />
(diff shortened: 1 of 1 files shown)


LucasF243:
it works on my side

installer/windows/build_installer.py
```diff
@@ -303,122 +303,153 @@ def create_wxs_file():
             Return='ignore'
         />
 
-        <CustomAction Id="ExtractMartools"
-            Directory="INSTALLDIR"
-            ExeCommand="powershell.exe -Command &quot;Expand-Archive '[MassaStationMartools]{MARTOOLS_ZIP}' -d '[INSTALLDIR]'&quot;"
+        <SetProperty Id="ExtractMartools" Value="&quot;powershell.exe&quot; -Command &quot;Expand-Archive '[MassaStationMartools]{MARTOOLS_ZIP}' -d '[INSTALLDIR]'&quot;" Before="ExtractMartools" Sequence="execute" />
```
Thykof, line 306: why Before="ExtractMartools" in an element with id ExtractMartools?

installer/windows/build_installer.py
```diff
@@ -303,122 +303,153 @@ def create_wxs_file():
             Return='ignore'
         />
 
-        <CustomAction Id="ExtractMartools"
-            Directory="INSTALLDIR"
```
Thykof, line 307: we don't need directory property anymore?

Thykof (approved):

installer/windows/build_installer.py
thomas-senechal: The "Directory" property isn't compatible with "BinaryKey" and "DllEntry" 
So we needed a workaround, the solution is simple, for "ExtractMartools", we don't need to be inside of the specific dir because the "Expand-Archive" command uses the Zip file we tell him to use, and unzip it where we tell it to unzip it

In case of things that need to be in a specific directory to work, for example Acrylic installation, we use a "Set-Location" (equivalent to `cd` in powershell) to move to the needed directory: https://github.com/massalabs/station/pull/1231/files/f39b276be6daa38f8f498c2450972819061ad758#diff-40085a0c3f8a1d74e1ec5324db618bbced740670ce3324856c05f60f09b4245bR346

gregLibert (approved):

PR #1234, opened 2023-10-18
Network storage cost

peterjah:


int/config/network.go
```diff
@@ -155,6 +136,27 @@ func (n *NetworkManager) Network() *NetworkInfos {
...
+
+	re := regexp.MustCompile(pattern)
+	matches := re.FindStringSubmatch(*status.Version)
+
+	//nolint:gomnd
+	if len(matches) != 2 {
+		return nil, fmt.Errorf("failed to parse node version from: %s", *status.Version)
+	}
+
+	return &matches[1], nil
```
gregLibert, line 157: Why not returning matches[1] instead ? this will make code easier to use/read since we won't have to dereference the string (version instead of *version)

gregLibert (approved):
One small remark. How should I test it ?

int/config/network.go
peterjah: the only reason i do this it do be able to return `nil , fmt.Errorf("....` but yes maybe we dont care and i can return `"", fmt.Error` instead, i dont know what is the good practice here

peterjah:
> One small remark. How should I test it ?

For now this only impact webOnchain, so you can try to deploy a website on testnet.
But i already spend time to adapt the robot tests so everything pass on tesnet26. (it is testing bunch of stuff like deploy sc and WoC).

int/config/network.go
gregLibert: The good practice is to return the default value of the type, here "".

PR #1235, opened 2023-10-18
fix error formating with nil pointers

Thykof:
Encountered this while working on https://github.com/massalabs/station-massa-wallet/pull/778

## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

### DIFF
--- a/pkg/node/events.go
+++ b/pkg/node/events.go
@@ -130,6 +130,5 @@ func ListenEvents(
 		}
 	}
 
-	// Handle the case where one or more pointers are nil
 	return nil, fmt.Errorf("no events found in the given time interval (%d seconds)", maxWaitingTimeInSeconds)
 }
--- a/pkg/node/events.go
+++ b/pkg/node/events.go
@@ -115,7 +115,7 @@ func ListenEvents(
 		events, err := Events(client, start, end, emitter, caller, operationID)
 		if err != nil {
 			return nil,
-				fmt.Errorf("fetching events for: opId %s, caller %s, emitter %s: %w", *operationID, *caller, *emitter, err)
+				fmt.Errorf("fetching events %w", err)
 		}
 
 		for _, event := range events {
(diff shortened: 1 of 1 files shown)


pkg/node/events.go
```diff
@@ -130,6 +130,28 @@ func ListenEvents(
...
+
+		if emitter == nil {
+			err += "emitter is nil, "
+		}
+
+		return nil, fmt.Errorf(err)
+	}
+
+	return nil, fmt.Errorf(
+		"timeout while listening events for: opId %s, caller %s, emitter %s", *operationID, *caller, *emitter)
```
peterjah, line 156: not very conviced by this function. maybe its normal that these pointers are nil in case of error. in that case just remove them from the error message

api/test/robot_tests/cmd/keywords.resource
```diff
@@ -16,8 +16,6 @@ Suite Setup
     ${GITHUB_ACTIONS}    Get Environment Variable    GITHUB_ACTIONS    false
     IF    "${GITHUB_ACTIONS}" == "true"
         Log To Console    We are in the CI
-        Delete Wallet Plugin
-        Install Wallet Plugin
```
peterjah, line 20: why are you deleting this?

peterjah: the wallet needs to be installed

Thykof: my bad

pkg/node/events.go
Thykof: these pointers can be nil, yes, the goal here is to prevent panic while formatting the error.

peterjah (approved):

pkg/node/events.go
```diff
@@ -130,6 +130,6 @@ func ListenEvents(
 		}
 	}
 
-	return nil,
-		fmt.Errorf("listening events for: opId %s, caller %s, emitter %s: Timeout", *operationID, *caller, *emitter)
+	// Handle the case where one or more pointers are nil
```
Thykof, line 133:
```suggestion
```

PR #1244, opened 2023-10-25
Add tos to linux installer

thomas-senechal:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**


installer/deb/create_deb.sh
```diff
@@ -77,22 +77,26 @@ main() {
     mkdir -p $BUILD_DIR/usr/share/pixmaps || fatal "failed to create $BUILD_DIR/usr/share/pixmaps"
     cp $TMP_DIR/usr/local/share/pixmaps/MassaStation.png $BUILD_DIR/usr/share/pixmaps || fatal "failed to copy MassaStation.png to $BUILD_DIR/usr/share/pixmaps"
 
+    mkdir -p $BUILD_DIR/usr/share/doc/massastation || fatal "failed to create $BUILD_DIR/usr/share/doc/massastation"
```
Thykof, line 80: because you do computation in the fatal string, is there a risk of a panic without any error message? Imagine you make a mistake in `fatal "failed to copy MassaStation.png to $BUILD_DIR/usr/share/pixmaps"`, for example the variable BUILD_DIR doesn't exist, what would be the error? Maybe it's safer to put only a simple string as `fatal` argument

Thykof (approved):
LGTM

installer/deb/create_deb.sh
thomas-senechal: In shell scripting, if a variable does not exists, it is replaced by nothing and no error is raised
So with you example, we would endup with `failed to copy MassaStation.png to /usr/share/pixmaps`
But also, `$BUILD_DIR` is set as a global variable at the beginning of the script, so that should never happen

And, anyway, this script is for building the deb package so if there were small errors like that, it should be fine

Thykof: ok!

installer/deb/create_deb.sh
```diff
@@ -77,22 +77,26 @@ main() {
...
     MassaStation is a secured gateway to the Massa blockchain. This application provides a user-friendly way to access, use and build on the Massa blockchain while keeping you safe from the dangers of the internet.
-Depends: curl, iproute2
+Depends: curl, iproute2, debconf (>= 0.5) | debconf-2.0
 Recommends: libnss3-tools
 EOF
 
-    cp deb/scripts/post* $BUILD_DIR/DEBIAN
-    DEB_NAME=massastation_$PKGVERSION\_amd64.deb
+    cp deb/scripts/* $BUILD_DIR/DEBIAN/ || fatal "failed to copy installer scripts to $BUILD_DIR/DEBIAN"
+    cp deb/templates $BUILD_DIR/DEBIAN/templates || fatal "failed to copy templates to $BUILD_DIR/DEBIAN"
```
peterjah, line 97: `deb/templates` does not exist. forgot to commit ?
if its a directory i think you need `cp -r`

peterjah (approved):

peterjah (approved):

gregLibert (approved):
It works with a couple of observations:
- The HTTP link for the Terms of Service isn't yet operational.
- I haven't verified if installing a new version of MS requires re-acceptance of the ToS.

thomas-senechal:
> * The HTTP link for the Terms of Service isn't yet operational.

Will work once merged

> * I haven't verified if installing a new version of MS requires re-acceptance of the ToS.

When the user accepts it the first time, we create an empty file, so next times the installer is ran (for example, for an update), we check if this file exists or not:
https://github.com/massalabs/station/pull/1244/files#diff-4bc1993c44b0052596d7cbce5d4d3082232197d6c3fa2437af8385ca92d5d860R39-R45

If it already exists, that means user already agreed so we don't need to ask again

gregLibert:
> > * The HTTP link for the Terms of Service isn't yet operational.
> 
> Will work once merged
> 
> > * I haven't verified if installing a new version of MS requires re-acceptance of the ToS.
> 
> When the user accepts it the first time, we create an empty file, so next times the installer is ran (for example, for an update), we check if this file exists or not: https://github.com/massalabs/station/pull/1244/files#diff-4bc1993c44b0052596d7cbce5d4d3082232197d6c3fa2437af8385ca92d5d860R39-R45
> 
> If it already exists, that means user already agreed so we don't need to ask again

I know, but I haven't tested it :-)

PR #1264, opened 2023-11-10
web-on-chain: improve and adjust for v27

Thykof:
- [x] add web on chain description in signature
- [x] increase default mas gas to 1 000 000 000 for web on chain to work on v27
- [x] rename some variables
- [x] keep same chunk size for web on chain

## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

### DIFF
--- a/pkg/onchain/website/uploader.go
+++ b/pkg/onchain/website/uploader.go
@@ -20,9 +20,9 @@ import (
 var content embed.FS
 
 const (
-	chunkSize = 260000
-	nbChunkKey  = "NB_CHUNKS"
-	ownerKey    = "OWNER"
+	chunkSize  = 260000
+	nbChunkKey = "NB_CHUNKS"
+	ownerKey   = "OWNER"
 )
 
 //nolint:funlen,cyclop
--- a/pkg/onchain/website/uploader.go
+++ b/pkg/onchain/website/uploader.go
@@ -20,7 +20,7 @@ import (
 var content embed.FS
 
 const (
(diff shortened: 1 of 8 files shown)


Thykof:
I tested, it doesn't work because of https://github.com/massalabs/station-massa-wallet/issues/845

but when I don't use correlation id (set correlation id manually in the code to "" in the /sign API call), it works for flappy and laby websites.

Thykof:
tested on securenet, it works, for webonchain, better put some fees (1000nanoMAS) on the appendBytesToWebsite.

pkg/onchain/website/uploader.go
```diff
@@ -20,7 +20,7 @@ import (
 var content embed.FS
 
 const (
-	blockLength = 260000
+	blockLength = 78000
```
peterjah, line 23: why this value ? the new blockSize is 300kb
puting 78kb here will make website upload very painfull :/ (lot of chunks to upload)

peterjah (changes requested):

pkg/onchain/website/uploader.go
Thykof: https://github.com/massalabs/station/issues/1263#issuecomment-1806512678

Thykof: how to chose the value then?

thomas-senechal (approved):

PR #1270, opened 2023-11-16
rename struct Wallet into Account

Thykof:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

### DIFF
--- a/pkg/my/domain.go
+++ b/pkg/my/domain.go
@@ -29,7 +29,7 @@ func Domains(config config.NetworkInfos, client *node.Client, nickname string) (
 
 	acc, err := wallet.Fetch(nickname)
 	if err != nil {
-		return nil, fmt.Errorf("loading wallet '%s': %w", nickname, err)
+		return nil, fmt.Errorf("loading account '%s': %w", nickname, err)
 	}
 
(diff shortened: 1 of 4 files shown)


peterjah (approved):

pkg/my/domain.go
```diff
@@ -27,18 +27,18 @@ and returns it as an array of strings.
 func Domains(config config.NetworkInfos, client *node.Client, nickname string) ([]string, error) {
 	const ownedPrefix = "owned"
 
-	wallet, err := wallet.Fetch(nickname)
+	acc, err := wallet.Fetch(nickname)
 	if err != nil {
 		return nil, fmt.Errorf("loading wallet '%s': %w", nickname, err)
```
thomas-senechal, line 32: I guess that if we start renaming "wallet" to "account" (where needed), we need to update logs/errors too ? 
```suggestion
		return nil, fmt.Errorf("loading account '%s': %w", nickname, err)
```

thomas-senechal (approved):
Just one small comment, otherwise it's good to me

PR #1273, opened 2023-11-16
1269 read only execute sc

Thykof:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**


pkg/onchain/sc.go
```diff
@@ -119,9 +119,21 @@ func DeploySC(client *node.Client,
 	signer signer.Signer,
 	description string,
 ) (*sendOperation.OperationResponse, []node.Event, error) {
+	// Calibrate max_gas
+	if maxGas == sendOperation.DefaultGasLimitExecuteSC || maxGas == 0 {
```
peterjah, line 123: this condition is strange, if we dont want to use DefaultGasLimitExecuteSC, we should delete it and do the estimation only when maxgas is not provided

thomas-senechal (approved):

peterjah (approved):

peterjah (changes requested):
from the swagger file:
```
              maxGas:
                $ref: "#/definitions/Amount"
                description: Maximum number of gas unit that a operation will be able consume.
                default: 700000000 # DefaultGasLimit
```
  
```
  Amount:
    description: Amount in nanoMassa.
    type: string
    x-nullable: false
```

    I think there is better to do here.
    I suggest to allow amount to be nullable here, and if null, do the estimation.
    If we don't want to change the behaviour everywhere, you can create a new swagger type 'nullableAmount' by instance...

peterjah (approved):

PR #1274, opened 2023-11-17
Massa v27 compatible

Thykof:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**


pkg/node/sendoperation/sendoperation.go
```diff
@@ -228,35 +233,35 @@ func DecodeOperationType(data []byte) (uint64, error) {
...
-func StorageCostForEntry(nodeVersion string, keyByteLengh, valueByteLenght int) (int, error) {
+func StorageCostForEntry(nodeVersion string, keyByteLength, valueByteLength int) (int, error) {
 	versionFloat, err := strconv.ParseFloat(nodeVersion, 64)
 	if err != nil {
-		return 0, fmt.Errorf("failed to parse nodeversion %s: %w", nodeVersion, err)
+		return 0, fmt.Errorf("failed to parse node version %s: %w", nodeVersion, err)
 	}
 
 	//nolint:gomnd
 	if versionFloat < 26 {
```
Thykof, line 243: @peterjah should we keep this code? because the order of parameter for event rpc calls are breaking change, so use can't use massa station on a network before v26 if we merge this breaking change.

peterjah: yes we can remove code for version <26

peterjah (approved):

PR #1275, opened 2023-11-17
EXPE: v27,  gas estimator, securenet

Thykof:
I opened this PR to test on the securenet27 the read only call and read only bytecode.

## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**


Thykof:
I successfully deployed a laby website

pkg/node/sendoperation/readonlycall.go
```diff
@@ -36,7 +36,7 @@ func EstimateGasCostCallSC(
 
 	estimatedGasCost := uint64(result.GasCost)
 
-	return estimatedGasCost, nil
+	return estimatedGasCost + DefaultInstalCreationCost, nil
```
Thykof, line 39: warning here, the value can go beyond the maximum value allowed for max gas, we should do something like max(estimatedGasCost + DefaultInstalCreationCost, MaxGasAllowedCallSC)

PR #1281, opened 2023-12-04
Sign and notarize macos app and installer

Thykof:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

### DIFF
--- a/installer/macos/create_pkg.sh
+++ b/installer/macos/create_pkg.sh
@@ -2,13 +2,21 @@
 
 # This script generates a .pkg file for the installation of MassaStation on Mac OS.
 
+# This script can be used in two contexts:
+# - in the CI,
+# - in local,
+# that's why there are some if/else statements.
+
 set -e
 
 PKGVERSION=dev
 ARCH=$1
 
-MASSASTATION_INSTALLER_NAME=MassaStation.app
+MASSASTATION_APPLICATION_NAME=MassaStation.app
 MASSASTATION_BINARY_NAME=massastation
+
+APPLE_DEVELOPER_ID_APPLICATION=$2
+APPLE_DEVELOPER_ID_INSTALLER=$3
 
 HOMEBREW_INSTALL_SCRIPT_URL=https://raw.githubusercontent.com/massalabs/homebrew.sh/master/homebrew-3.3.sh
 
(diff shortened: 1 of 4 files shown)


installer/macos/create_pkg.sh
```diff
@@ -7,9 +7,12 @@ set -e
...
-MASSASTATION_INSTALLER_NAME=MassaStation.app
+MASSASTATION_APPLICATION_NAME=MassaStation.app
 MASSASTATION_BINARY_NAME=massastation
 
+APPLE_DEVELOPER_ID_APPLICATION=$2
+APPLE_DEVELOPER_ID_INSTALLER=$3
+
 HOMEBREW_INSTALL_SCRIPT_URL=https://raw.githubusercontent.com/massalabs/homebrew.sh/master/homebrew-3.3.sh
 
 LICENSE_FILE_NAME=MassaStation_ToS.rtf
```
thomas-senechal, line 18: I can't comment on the `display_usage` function so I'm commenting here. 
I think we need to update it to add the App Dev ID and Installer Dev ID

thomas-senechal (changes requested):
Did you tried to run the `create_pkg.sh` script from your computer without the Dev IDs ? 
What happens in that case ? 

If packaging fails or if the installer does not work, I think we need to add an handling of building with and without the Dev IDs. Otherwise, we won't be able to build from our computers and will require to push, wait for 15min, to then check if it's working. And it's not a viable way of working IMO

Thykof:
> Did you tried to run the `create_pkg.sh` script from your computer without the Dev IDs ? What happens in that case ?
> 
> If packaging fails or if the installer does not work, I think we need to add an handling of building with and without the Dev IDs. Otherwise, we won't be able to build from our computers and will require to push, wait for 15min, to then check if it's working. And it's not a viable way of working IMO

I modifier the script, you can now run `./macos/create_pkg.sh amd64 "Developer ID Application: Massa Labs (9X8NWA3JHQ)" "Developer ID Installer: Massa Labs (9X8NWA3JHQ)"` or `./macos/create_pkg.sh amd64`

installer/macos/create_pkg.sh
```diff
@@ -38,20 +48,31 @@ install_massastation_build_dependencies() {
...
+    fyne package -icon ../../int/systray/embedded/logo.png -name MassaStation -appID com.massalabs.massastation -src ../cmd/massastation || fatal "fyne package failed for $MASSASTATION_APPLICATION_NAME"
+    chmod +x $MASSASTATION_APPLICATION_NAME || fatal "failed to chmod $MASSASTATION_APPLICATION_NAME"
 }
 
 # Build the package using pkgbuild.
 package() {
-    pkgbuild --component $MASSASTATION_INSTALLER_NAME --identifier com.massalabs.massastation --version $PKGVERSION \
+    # sign the application if we have a developer id
+    if [[ -n "$APPLE_DEVELOPER_ID_APPLICATION" ]]; then
+        codesign --force --options runtime --sign "$APPLE_DEVELOPER_ID_APPLICATION" $MASSASTATION_APPLICATION_NAME
```
thomas-senechal, line 63: Why `--force` ? 🤔 

(non blocking)

thomas-senechal (approved):
Good job ! 
Only a non blocking comment

gregLibert (approved):

installer/macos/create_pkg.sh
Thykof: to replace any signature already there, it fails if there is already a signature and we don't use the --force option. However, there should not be a signature already but let's be sure.

PR #1293, opened 2023-12-11
1286 remove acrylic dns from windows installer

thomas-senechal:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

# QA Review
1. [Uninstall MS](https://docs.massa.net/docs/massaStation/uninstall#windows)
2. Download the installer from https://github.com/massalabs/station/actions/runs/7169232980?pr=1293
3. Install MS
4. Run MS, it should work fine (except WoC since `*.massa` is disable)
5. Open file explorer and copy paste the following path:
    ```
    %windir%\System32\drivers\etc\
    ```
6. Open the `hosts` file
7. You should see a line with `127.0.0.1 station.massa` 
8. You can close the file, but not the file explorer
9. Uninstall MS again
10. Reopen the `hosts` file
11. You should see no line mentionning `station.massa`


peterjah (approved):

gregLibert (approved):
A side note:

We might consider renaming our installer files, using 'installer' as a prefix. This would ensure they appear first in the list of artifacts and enhance readability.

What do you think? Of course, this change should be addressed in a separate issue and would require synchronization with the website, among other things.

thomas-senechal:
> A side note:
> 
> We might consider renaming our installer files, using 'installer' as a prefix. This would ensure they appear first in the list of artifacts and enhance readability.
> 
> What do you think? Of course, this change should be addressed in a separate issue and would require synchronization with the website, among other things.

If you are talking about the release artifacts, I'm not sure it's really important since most people are using the website to download the installers.

But this change shouldn't take much time and should be better for users downloading throught GitHub releases

PR #1297, opened 2023-12-18
add chain id

Thykof:
related to https://github.com/massalabs/massa/issues/4583

## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**


int/config/network.go
```diff
@@ -166,16 +135,33 @@ func (n *NetworkManager) SwitchNetwork(selectedNetworkStr string) error {
...
+
+	version, err := versionFromStatus(status)
 	if err != nil {
 		return fmt.Errorf("getting network version: %w", err)
 	}
 
+	// compare chain id from node status with chain id from config
+	nodeChainID := uint64(*status.ChainID)
+	if nodeChainID != cfg.ChainID {
+		logger.Errorf("chain id mismatch: %d != %d", nodeChainID, cfg.ChainID)
```
Thykof, line 154: should we exit here?

```suggestion
		logger.Fatalf("chain id mismatch: %d != %d", nodeChainID, cfg.ChainID)
```

thomas-senechal: I guess we shouldn't exit. It's a node issue. 
But we could display an error (or a warning) to tell user that there is something weird with the connected node and so, that we recommand switching node.

Maybe that should be done in another issue/PR as I don't think it is a priority

thomas-senechal (approved):

damip:
CI seems to fail

Thykof:
> CI seems to fail

CI fails because buildnet is not handling chain id yet

peterjah (approved):

PR #1299, opened 2023-12-18
nss functions: don't block on error

Thykof:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

### DIFF
--- a/pkg/nss/manager.go
+++ b/pkg/nss/manager.go
@@ -41,13 +41,15 @@ func NewManager(dbPath []string, certutil CertUtilServicer, logger Logger) *Mana
 
 // executeOnPaths executes the given operation on each NSS database path.
 func (m *Manager) executeOnPaths(operation func(path string) error) error {
+	var totalErr error
+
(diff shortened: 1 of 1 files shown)


pkg/nss/manager.go
```diff
@@ -41,13 +41,15 @@ func NewManager(dbPath []string, certutil CertUtilServicer, logger Logger) *Mana
 
 // executeOnPaths executes the given operation on each NSS database path.
 func (m *Manager) executeOnPaths(operation func(path string) error) error {
+	var totalErr error
```
gregLibert, line 44: The variable name `totalErr` isn't very clear; it doesn't convey the purpose of the variable effectively. Since you're accumulating errors in this variable, a more descriptive name like `chainedErr` would be better to indicate its use for chaining errors.

gregLibert (changes requested):

gregLibert (approved):

thomas-senechal (approved):

PR #1309, opened 2024-01-03
add read only call sc endpoint

Thykof:
## TODO
- [x] add endpoint
- [x] add robot framework test

## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

## QA

```curl
curl --location 'http://station.massa/cmd/read-only/callsc' \
--header 'Content-Type: application/json' \
--data '{
  "nickname": "fullpower",
  "name": "rien",
  "at": "AU1MPDRXuR22mwYDFCeZUDgYjcTAF1co6xujx2X6ugoHeYeGY3B5",
  "args": "",
  "coins": "1000446777777"
}'
```


int/api/cmd/utils.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"strconv"
+
+	"github.com/go-openapi/runtime/middleware"
+	"github.com/massalabs/station/api/swagger/server/models"
+	"github.com/massalabs/station/api/swagger/server/restapi/operations"
+	sendOperation "github.com/massalabs/station/pkg/node/sendoperation"
+)
+
+// amountToUint64 converts fee to uint64.
```
Thykof, line 12:
```suggestion
// amountToUint64 converts the given amount to uint64.
```

int/api/cmd/utils.go
```diff
@@ -0,0 +1,126 @@
...
+			&models.Error{
+				Code:    errorInvalidArgs,
+				Message: "Error during amount conversion: " + err.Error(),
+			})
+	}
+
+	return amountString, nil
+}
+
+// CreateReadOnlyResult Converts an instance of sendOperation.ReadOnlyResult struct to  models.ReadOnlyResult struct
```
Thykof, line 50:
```suggestion
// CreateReadOnlyResult Converts an instance of sendOperation.ReadOnlyResult struct to models.ReadOnlyResult struct
```

int/api/cmd/read_only_call_sc.go
```diff
@@ -0,0 +1,69 @@
...
+		return operations.NewCmdReadOnlyCallSCBadRequest().WithPayload(
+			&models.Error{
+				Code:    errorInvalidNickname,
+				Message: "Error during wallet fetch: " + err.Error(),
+			})
+	}
+
+	if params.Body.Fee == "" {
+		params.Body.Fee = "0"
+	}
```
thomas-senechal, line 45: What about setting the default value directly in the swagger file ? 
That would be simpler than setting default values in the code like that 🤔

int/api/cmd/read_only_call_sc.go
```diff
@@ -0,0 +1,69 @@
...
+	if errResponse != nil {
+		return errResponse
+	}
+
+	result, err := sendOperation.ReadOnlyCallSC(
+		params.Body.At,
+		params.Body.Name,
+		args,
+		coins,
+		string(params.Body.Fee),
```
thomas-senechal, line 57: If `Fee` is already of type `string`, why make it as a string again ? 🤔

Thykof: Fee is of type Amount which is type string but golang don't want implicit cast

int/api/cmd/read_only_call_sc.go
Thykof: you're right

Thykof: but it doesn't work

PR #1314, opened 2024-01-08
Update go and linter in ci

thomas-senechal:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**


.github/actions/install/action.yml
```diff
@@ -17,11 +17,10 @@ runs:
         sudo apt update
         sudo apt install -y libgl1-mesa-dev xorg-dev gcc-mingw-w64-x86-64 p7zip
 
-    - uses: actions/setup-go@v4
+    - uses: actions/setup-go@v5
       with:
-        go-version: 1.20.6
-        cache: true
```
Thykof, line 23: we don't need `cache: true`?

thomas-senechal: No
It's on by default since v3

Thykof (approved):

PR #1316, opened 2024-01-08
917 refactor front interfaces

Thykof:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

### DIFF
--- a/web/massastation/src/main.tsx
+++ b/web/massastation/src/main.tsx
@@ -11,6 +11,7 @@ import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 
 import '@massalabs/react-ui-kit/src/global.css';
 import './index.css';
+import { IconContext } from 'react-icons/lib';
 
 import { ENV } from './const/env/env';
 import {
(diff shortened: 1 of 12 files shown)


thomas-senechal (approved):

web/massastation/src/main.tsx
```diff
@@ -56,8 +57,10 @@ const router = createBrowserRouter(
 
 ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
   <React.StrictMode>
-    <QueryClientProvider client={queryClient}>
-      <RouterProvider router={router} fallbackElement={<Error />} />
-    </QueryClientProvider>
+    <IconContext.Provider value={{ size: '32' }}>
```
pivilartisant, line 60: quick question: why did you add back the icon context provider ?

Thykof: because the size was too small when I use npm link, so this way we explicitly set the size to 32 of all icons.

PR #1320, opened 2024-01-09
network config in config yaml file

Thykof:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

### DIFF
--- a/int/config/network.go
+++ b/int/config/network.go
@@ -1,8 +1,9 @@
 package config
 
 import (
-	"embed"
 	"fmt"
+	"os"
+	"path"
 	"regexp"
 	"sync"
 
@@ -12,7 +13,13 @@ import (
 )
 
 const (
-	MassaStationURL = "station.massa"
+	MassaStationURL   = "station.massa"
+	networkConfigFile = "config_network.yaml"
+	mainnetRPC        = "https://mainnet.massa.net/api/v2"
+	mainnetChainID    = 77658377
+	buildnetRPC       = "https://buildnet.massa.net/api/v2"
+	buildnetChainID   = 77658366
+	permissionUrwGrOr = 0o644
 )
 
 type NetworkInfos struct {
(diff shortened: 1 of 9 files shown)


peterjah (approved):

int/config/network.go
```diff
@@ -167,21 +174,21 @@ func (n *NetworkManager) SwitchNetwork(selectedNetworkStr string) error {
 	return nil
 }
 
-//nolint:typecheck,nolintlint
-//go:embed config_network.yaml
-var configData embed.FS
```
thomas-senechal, line 172: Shouldn't we embed a default yaml file ? 
That would avoid us having to remake it one from the code and make it simpler to modify since we would easily know where the yaml file is in the code.

int/config/network.go
```diff
@@ -167,21 +174,21 @@ func (n *NetworkManager) SwitchNetwork(selectedNetworkStr string) error {
...
-// LoadConfig loads network configurations from an embedded YAML file.
+// LoadConfig loads network configurations from YAML file.
 // Returns the loaded network configurations and any error encountered during loading.
-func LoadConfig() (map[string]NetworkConfig, error) {
-	var networksData map[string]NetworkConfig
+func LoadConfig(configDir string) (map[string]NetworkConfig, error) {
+	networkConfigPath := path.Join(configDir, networkConfigFile)
+	if _, err := os.Stat(networkConfigPath); os.IsNotExist(err) {
+		createDefaultConfig(networkConfigPath)
+	}
```
thomas-senechal, line 183: What if it's not a `os.IsNotExist` error?

Thykof: good point, I need to handle this error

int/config/network.go
Thykof: the file is created if not found

Thykof: well, I'm not convinced because the code will create a file if it doesn't exists

thomas-senechal: As discussed during our daily, we might want to add comments in the file to warn users that the default nodes (Massa Mainnet and Buildnet) shouldn't be modified by users if they don't want to break MS.
Also, we might want to add a short comment at the beginning explaining what this config file is and other stuff like that

I'm unsure it is possible to add comments from code (at least using yaml marshall), so maybe that would force us to have a default yaml file embedded

thomas-senechal (approved):

int/config/network.go
Thykof: @SlnPons does the doc is sufficient or do we need to do what Thomas suggest (adding comments in the .yaml file)?

PR #1323, opened 2024-01-17
Add docs link in station homepage

pivilartisant:
open source contribution: 
add link to station homepage so users have a quick access 

Screenshot 2024-01-17 at 11 00 51


SlnPons:
@pivilartisant thanks a lot for this opportunistic PR. Love the spirit!! 
It's a good idea but I would have redirected to MS doc instead of the general doc and more specifically to the "how to build a plugin" one. But at this stage of the product, this document is still unsatisfactory IMO. 
This is why I would prefer close this PR for now and wait for a proper doc to exist to put it in the home page :)

PR #1341, opened 2024-04-08
1340 minimal fees

Thykof:
draft because maybe we want to do that in the wallet plugin...


Thykof:
close because the logic is done in the wallet plugin, and following PR here https://github.com/massalabs/station/pull/1344 to keep some refacto

PR #1351, opened 2024-04-22
Enable dashboard plugin update

pivilartisant:
Changes: 

Plugin Wallet is now a component hosted in station repo (not ui-kit)
Update Plugin logic is contained in a custom hook and called in two places: dashboard and plugin store 

These changes enables the user to directly update MW plugin in the dashboard ui without having to navigate to the plugin store

### DIFF
--- a/web/massastation/src/pages/Index/Dashboard/MassaWallet.tsx
+++ b/web/massastation/src/pages/Index/Dashboard/MassaWallet.tsx
@@ -2,7 +2,7 @@
 // @ts-ignore
 import { Button, Spinner } from '@massalabs/react-ui-kit';
 import Intl from '@/i18n/i18n';
-import { useEffect, useState } from 'react';
+import { useState } from 'react';
 import { FiArrowUpRight, FiRefreshCw } from 'react-icons/fi';
 import { WalletStates } from '../DashboardStation';
 import { motion } from 'framer-motion';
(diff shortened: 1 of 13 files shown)


Thykof (approved):

web/massastation/src/pages/Index/DashboardStation.tsx
```diff
@@ -27,20 +35,50 @@ export function DashboardStation(props: IDashboardStationProps) {
...
+
+  useEffect(() => {
+    if (pluginWalletIsInstalled && !isUpdatable) {
+      setWalletState(WalletStates.Active);
+    } else if (pluginWalletIsInstalled && isUpdatable) {
+      setWalletState(WalletStates.Updateable);
+    } else {
+      setWalletState(WalletStates.Inactive);
+    }
+  }, [isUpdatable, pluginWalletIsInstalled]);
```
Thykof, line 59: try without a use effect and without a use state for walletState, just a const might work

Thykof (changes requested):

web/massastation/src/pages/Index/DashboardStation.tsx
pivilartisant: Unfortunately we need a re-render for the wallet even for this initial state... I tested logging the current states and the result was ```undefined```, ```inactive``` and finally ```active``` if you want we can sync for 5 mins to discuss this issue

web/massastation/src/pages/Index/Dashboard/MassaWallet.tsx
```diff
@@ -2,45 +2,56 @@
...
+  isHovered?: boolean;
 }
 
 export function ActivePlugin(props: MSPluginProps) {
-  const { title, iconActive, onClickActive } = props;
+  const { title, onClickActive, isHovered } = props;
+
+  useEffect(() => {
+    console.log('isHovered', isHovered);
+  }, [isHovered]);
```
Thykof:
```suggestion
```

PR #1352, opened 2024-04-23
Fix end2end wallet test

Thykof:
we need to release a version of the wallet after merging https://github.com/massalabs/station-massa-wallet/pull/919 and update the robot test to install this version

here you can change the version https://github.com/massalabs/station/blob/10b8154b29843d5bc4ca746639272320496f014a/api/test/robot_tests/cmd/keywords.resource#L10

### DIFF
--- a/api/swagger/server/restapi/resource/swagger.yml
+++ b/api/swagger/server/restapi/resource/swagger.yml
@@ -124,8 +124,9 @@ paths:
                 $ref: "#/definitions/Amount"
                 description: Set the coin amount that will be transferred to the smartContract.
               fee:
-                $ref: "#/definitions/Amount"
-                description: Set the fee amount that will be given to the block creator.
+                type : string
+                description: Set the fee amount (in massa) that will be given to the block creator.
+                default: "0"
       produces:
         - application/json
       responses:
--- a/api/test/robot_tests/cmd/cmd.robot
+++ b/api/test/robot_tests/cmd/cmd.robot
@@ -16,7 +16,7 @@ POST /cmd/read-only/executesc
     ${data}=    Create Dictionary
     ...    nickname=${WALLET_NICKNAME}
     ...    coins=3000000000
-    ...    fee=1000
+    ...    fee=0.01
     ${file}=    Create Dictionary    bytecode=${sc}
     ${response}=    POST    ${API_URL}/cmd/read-only/executesc    data=${data}    files=${file}    expected_status=any
     Log To Console    json response: ${response.json()}    # Print the response content to the test log for debugging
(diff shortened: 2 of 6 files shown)


api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -124,8 +124,9 @@ paths:
                 $ref: "#/definitions/Amount"
                 description: Set the coin amount that will be transferred to the smartContract.
               fee:
-                $ref: "#/definitions/Amount"
-                description: Set the fee amount that will be given to the block creator.
+                type : string
+                description: Set the fee amount (in massa) that will be given to the block creator.
+                default: "0"
```
Thykof, line 129: here we fix the fee parameter, before it was a Amount in nonamassa but the endpoint expect to have the value in MAS. It was not failed the tests because 1000 nano mas or mas, it passes.

also, we fix the documentation

api/test/robot_tests/cmd/cmd.robot
Thykof, line 1: for read only endpoints, the fee parameter is in MAS and for exectu sc, call sc endpoints the fee parameter is in nanomassa

api/test/robot_tests/cmd/keywords.resource
```diff
@@ -7,7 +7,7 @@ Resource            ../keywords.resource
 Resource            ../variables.resource
 
 *** Variables ***
-${WALLET_PLUGIN_VERSION}       v0.3.0
+${WALLET_PLUGIN_VERSION}       v0.3.5
```
Thykof, line 10: v0.3.5 is a prerelease, nothing change for the end user, only the env prompter is updated with the minimal fees, here we use this version to be able to submit op in the buildnet.

api/test/robot_tests/event_manager/event_manager.robot
```diff
@@ -26,11 +26,13 @@ GET /events/{str}/{caller} from called function
...
     ${handle}=    Async Run    GET
     ...    ${API_URL}/events/${expected_event}/${WALLET_ADDR}
-    ...    expected_status=${STATUS_OK}
+    ...    expected_status=any
 
     Generate Event    ${randomID}
 
     ${response}=    Async Get    ${handle}
+    Log To Console    json response: ${response.json()}    # Print the response content to the test log for debugging
+    Should Be Equal As Integers    ${response.status_code}    ${STATUS_OK}    # Assert the status code is 200 OK
```
Thykof, line 35: we simply add some logs to easy debug

int/api/cmd/read_only_call_sc.go
Thykof, line 1: fee was before the Amount, for it's a pointer to a string, because of the change in the swagger

api/test/robot_tests/cmd/cmd.robot
thomas-senechal: That's weird, we should use the same unit eve rywhere, so I guess, nMAS. 

Maybe we should open an issue on that if you feel it's out of the scope of this PR?

PR #1365, opened 2024-08-08
Add Delete Fn in Cache pkg

pivilartisant:
## Delivery Quality Checklist

- [x] **Breaking Changes in API:**

- [x] **Changelog:**

- [x] **Version Update Handling:**

- [x] **PR Dependency:**

- [x] **Documentation:**

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -80,7 +80,6 @@ func (c *Cache) Save(fileName string, content []byte) error {
 
 // Delete removes a file from the cache.
 func (c *Cache) Delete(fileName string) error {
-
 	fullPath, err := fullPath(fileName, c.ConfigDir)
 	if err != nil {
 		return fmt.Errorf("while reading cached file %s: %w", fileName, err)
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -87,6 +87,10 @@ func (c *Cache) Delete(fileName string) error {
 
 	err = os.Remove(fullPath)
 	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+
 		return fmt.Errorf("while deleting from cache: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/cache/cache.go
```diff
@@ -77,3 +77,19 @@ func (c *Cache) Save(fileName string, content []byte) error {
...
+func (c *Cache) Delete(fileName string) error {
+
+	fullPath, err := fullPath(fileName, c.ConfigDir)
+	if err != nil {
+		return fmt.Errorf("while reading cached file %s: %w", fileName, err)
+	}
+
+	err = os.Remove(fullPath)
+	if err != nil {
+		return fmt.Errorf("while deleting from cache: %v", err)
```
Thykof, line 91:
```suggestion
		return fmt.Errorf("while deleting from cache: %w", err)
```

pkg/cache/cache.go
```diff
@@ -77,3 +77,19 @@ func (c *Cache) Save(fileName string, content []byte) error {
 
 	return nil
 }
+
+// Delete removes a file from the cache.
+func (c *Cache) Delete(fileName string) error {
+
```
Thykof, line 83:
```suggestion
```

pkg/cache/cache.go
```diff
@@ -77,3 +77,18 @@ func (c *Cache) Save(fileName string, content []byte) error {
...
+func (c *Cache) Delete(fileName string) error {
+	fullPath, err := fullPath(fileName, c.ConfigDir)
+	if err != nil {
+		return fmt.Errorf("while reading cached file %s: %w", fileName, err)
+	}
+
+	err = os.Remove(fullPath)
+	if err != nil {
+		return fmt.Errorf("while deleting from cache: %w", err)
+	}
```
thomas-senechal, line 91: IMO, if file does not exist, we should treat it as success

```suggestion
	if err != nil {
		if os.IsNotExist(err) {
			return nil
		}

		return fmt.Errorf("while deleting from cache: %w", err)
	}
```

thomas-senechal (approved):

PR #1370, opened 2024-11-26
add deploy Sc

pivilartisant:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**


int/api/cmd/deploySC.go
```diff
@@ -13,6 +15,9 @@ import (
 	"github.com/massalabs/station/pkg/onchain"
 )
 
+//go:embed sc
+var content embed.FS
```
thomas-senechal, line 19: You can directly embed the SC like: 
```suggestion
//go:embed sc/deployer.wasm
var deployerSC []byte
```

Like that, you don't need to do the `content.ReadFile` line 31

pkg/onchain/datastore.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Add Uint8Array equivalent (assuming a single byte array with a 0 value)
+	buf.Write([]byte{1})
+
+	// Return the serialized data
+	return buf.Bytes(), nil
+}
```
pivilartisant, line 53:
```suggestion
func argsKey(offset int) ([]byte, error) {
	// Create a buffer to serialize data
	var buf bytes.Buffer

	// Add U64 (64-bit unsigned integer)
	err := binary.Write(&buf, binary.LittleEndian, uint64(offset+1))
	if err != nil {
		return nil, err
	}

	// Add Uint8Array equivalent (assuming a single byte array with a 0 value)
	buf.Write([]byte{0})

	// Return the serialized data
	return buf.Bytes(), nil
}

func coinsKey(offset int) ([]byte, error) {
	// Create a buffer to serialize data
	var buf bytes.Buffer

	// Add U64 (64-bit unsigned integer)
	err := binary.Write(&buf, binary.LittleEndian, uint64(offset+1))
	if err != nil {
		return nil, err
	}

	// Add Uint8Array equivalent (assuming a single byte array with a 0 value)
	buf.Write([]byte{1})

	// Return the serialized data
	return buf.Bytes(), nil
}
```

api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -207,54 +207,50 @@ paths:
...
+                description: Set the fee amount (in nanoMassa) that will be given to the block creator.
+                type : integer
+                minimum: 0
+                format: uint64
+                default : 0
+              parameters:
+                description: Base64 encoded parameters that will be sent along the smart contract.
+                type: string
+                default: ""
+              datastore:
```
peterjah, line 250: i dont think its needed

int/api/cmd/deploySC.go
```diff
@@ -45,22 +52,18 @@ func (d *deploySC) Handle(params operations.CmdDeploySCParams) middleware.Respon
...
+		sendoperation.MaxGasAllowedExecuteSC, // default
+		*params.Body.MaxCoins, // maxCoins 
+		*params.Body.Coins,    // smart contract deployment "fee"
+		sendoperation.DefaultExpiryInSlot,
+		parameters, 
+		smartContractByteCode,
+		deployerSCByteCode,
 		&signer.WalletPlugin{},
-		"",
+		"Deploying website",
```
peterjah, line 66: Deploying contract ?? 
Maybe you can add a "description" field to the endpoint to make it dynamic

pkg/node/sendoperation/sendoperation.go
```diff
@@ -105,22 +99,13 @@ func Call(
...
-			"batch": true, "chainId": ` + strconv.FormatUint(chainID, 10) + `}`
-	case operationBatch.CorrelationID != "":
-		content = descriptionLabel + description + `", "operation": "` + msgB64 + `",
-			"correlationId": "` + operationBatch.CorrelationID + `", "chainId": ` + strconv.FormatUint(chainID, 10) + `}`
-	default:
-		content = descriptionLabel + description + `",
-			"operation": "` + msgB64 + `", "chainId": ` + strconv.FormatUint(chainID, 10) + `}`
-	}
+	content = descriptionLabel + description + `",
+		"operation": "` + msgB64 + `", "chainId": ` + strconv.FormatUint(chainID, 10) + `}`
```
peterjah, line 108: This is very ugly ^^'
here is a chatGpt proposition:
```
	data := map[string]interface{}{
		"description": description,
		"operation":   msgB64,
		"chainId":     chainID,
	}

	content, err := json.Marshal(data)
	if err != nil {
		log.Fatalf("Error marshaling JSON: %v", err)
	}
	```

File: pkg/onchain/datastore.go [hunk 6]
```diff
@@ -0,0 +1,91 @@
+package onchain
+
+import (
+	"bytes"
+	"encoding/binary"
+
+	"github.com/massalabs/station/pkg/convert"
+)
+
+type DatastoreContract struct {
```
peterjah, line 10: maybe ContractDatastore ?

pkg/onchain/datastore.go
```diff
@@ -0,0 +1,91 @@
...
+	return args
+}
+
+/** 
+populateDatastore creates and serializes a datastore for the given contract.
+*/
+func populateDatastore(contract DatastoreContract) ([]byte, error) {
+	var datastore []datastoreEntry
+
+	// nomber of contracts to deploy
```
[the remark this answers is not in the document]
peterjah: maybe add comment to say taht for now we can only deploy 1 contract

pkg/onchain/datastore.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	return serializedDatastore, nil
+}
+
+// SerializeDatastore serializes the datastore into a []byte array.
+func SerializeDatastore(datastore []datastoreEntry) ([]byte, error) {
+	var buffer bytes.Buffer
+
+	buf := make([]byte, binary.MaxVarintLen64)
```
peterjah, line 66: as you use a go slice, i think you are not forced to declare the size. it should be dynamic

pkg/onchain/datastore.go
```diff
@@ -0,0 +1,91 @@
...
+	// Encode the number of key-value pairs
+	datastoreSize := uint64(len(datastore))
+	uDatastoreSize := binary.PutUvarint(buf, datastoreSize)
+
+	buffer.Write(buf[:uDatastoreSize])
```
peterjah, line 71: maybe buf.append(....) should work ?

pkg/onchain/datastore.go
pivilartisant: it's necessary "Slice: The size specifies the length. The capacity of the slice is
equal to its length. A second integer argument may be provided to
specify a different capacity; it must be no smaller than the
length. For example, make([]int, 0, 10) allocates an underlying array
of size 10 and returns a slice of length 0 and capacity 10 that is
backed by this underlying array." for maps it is not

pkg/onchain/datastore.go
pivilartisant: @thomas-senechal  what do u think ?

api/swagger/server/restapi/resource/swagger.yml
```diff
@@ -207,54 +207,50 @@ paths:
...
+                x-nullable: false
+              maxCoins:
+                description: Set the number of coins that will be sent along the deployment call.
+                type : integer
+                format: uint64
+                minimum: 0
+              coins:
+                description: Set the number of coins that will be sent along the deployment call.
+                type : integer
+                format: uint64
```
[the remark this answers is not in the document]
pivilartisant: so that would concern fee, maxCoins & coins fields ?

pkg/node/sendoperation/sendoperation.go
```diff
@@ -78,7 +76,8 @@ func Call(
 		return nil, err
 	}
 
-	content := createOperationContent(operationBatch, description, msgB64, chainID)
+	content := createOperationContent(description, msgB64, chainID)
+	logger.Infof("json content: %v",content)
```
peterjah, line 80: to delete

int/api/cmd/deploySC.go
```diff
@@ -22,7 +27,9 @@ type deploySC struct {
 }
 
 func (d *deploySC) Handle(params operations.CmdDeploySCParams) middleware.Responder {
-	file, err := io.ReadAll(params.SmartContract)
+	_smartContractBytes, err := base64.StdEncoding.DecodeString(params.Body.SmartContract)
+	smartContractReader := bytes.NewReader(_smartContractBytes)
+	smartContractByteCode, err := io.ReadAll(smartContractReader)
```
thomas-senechal, line 32: 2 things here:
- The first `err` is never checked
- I think it can be simplified as follow (but I didn't test it):
  ```suggestion
  	smartContractByteCode, err := base64.StdEncoding.DecodeString(params.Body.SmartContract)
  ```
  Because `base64.StdEncoding.DecodeString` already returns a `[]byte`, so there is no need to use `bytes.NewReader` and `io.ReadAll`

int/api/cmd/deploySC.go
```diff
@@ -31,11 +38,10 @@ func (d *deploySC) Handle(params operations.CmdDeploySCParams) middleware.Respon
...
 	}
-	/* All the pointers below cannot be null as the swagger hydrate
-	each one with their default value defined in swagger.yml,
-	if no values are provided for these parameters.
-	*/
-	decodedDatastore, err := base64.StdEncoding.DecodeString(*params.Datastore)
+
+	_parameters, err := base64.StdEncoding.DecodeString(params.Body.Parameters)
+	parameterReader := bytes.NewReader(_parameters)
+	parameters, err := io.ReadAll(parameterReader)
```
thomas-senechal, line 44: Same as before

pkg/onchain/datastore.go
thomas-senechal: I don't really understand TBH
If the question is about the PutUvarint, AFAIK, we need it to encode the datastoreSize to a byte array

PR #1374, opened 2025-01-10
fix readonly estimation

pivilartisant:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ on:
     inputs:
       tag_name:
         type: string
-        description: 'The tag name of the release without v prefix'
+        description: "The tag name of the release without v prefix"
 
 env:
   VERSION: ${{ inputs.tag_name }}
--- a/api/swagger/server/restapi/resource/swagger.yml
+++ b/api/swagger/server/restapi/resource/swagger.yml
@@ -128,7 +128,7 @@ paths:
                 $ref: "#/definitions/Amount"
                 description: Set the coin amount that will be transferred to the smartContract.
               fee:
-                type : string
+                type: string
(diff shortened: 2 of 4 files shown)


int/api/cmd/utils.go
```diff
@@ -115,7 +115,7 @@ func CreateReadOnlyResult(result sendOperation.ReadOnlyResult) models.ReadOnlyRe
 				Balance: &models.ChangeSet{
 					Set: oldChange.Update.Balance.Set,
 				},
-				Bytecode:  oldChange.Update.Bytecode,
+				Bytecode:  oldChange.Update.Bytecode.(string),
```
peterjah, line 118: I think this is not good.
Here you handle only the case where Bytecode is a string, which is the case when there is no bytecode change and the value is "Keep"
If you want to handle the case where Bytecode is a ChangeSet object, you must detect the type here as following:
		switch v := oldChange.Update.Bytecode.(type) {
		case string:
			newChange.Update.Bytecode = v
		case models.ChangeSet:
			newChange.Update.Bytecode = v
		default:
			newChange.Update.Bytecode = nil
		}

you also need to modify the type of newChange.Update.Bytecode from string to interface{

peterjah (changes requested):

int/api/cmd/utils.go
peterjah: you can achieve this by modify the api yml file directly

peterjah: line 1063 type: string -> type: object

damip:
@thomas-senechal any news on this ?
also needs priorization

PR #1379, opened 2025-01-16
add datastore handling function for detecting deploysc op

fleandrei:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**

### DIFF
--- a/pkg/onchain/datastore.go
+++ b/pkg/onchain/datastore.go
@@ -73,6 +73,7 @@ func getArgsKey() []byte {
 // getCoinsKey returns the key in the datastore for the amount of MAS to be sent to the deployed contract.
 func getCoinsKey() []byte {
 	lengthPrefix := convert.U32ToBytes(byteArrayLengthPrefix)
+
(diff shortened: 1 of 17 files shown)


pkg/onchain/datastore.go
```diff
@@ -54,8 +55,71 @@ func populateDatastore(contract ContractDatastore) ([]byte, error) {
...
+
+	var err error
+
+	contract.Coins, err = convert.BytesToU64(datastore[3].Value)
+	if err != nil {
+		return ContractDatastore{}, false
+	}
+
+	return contract, true
+}
```
peterjah, line 106: What we usually to is return  data, error  for function return value.
When error = nil its equivalent to your success=true
when error != nil we can get the error message.

what do you think?

peterjah: then in the calling function we can check the error to see whats to do

peterjah (approved):

PR #1382, opened 2025-02-12
rework install plugin from zip

peterjah:


pkg/node/sendoperation/sendoperation.go
```diff
@@ -222,8 +222,6 @@ func DecodeMessage64(msgB64 string) ([]byte, uint64, uint64, error) {
 }
 
 // DecodeOperationType decodes a byte slice to retrieve the operation ID.
-// Can only be used on operation specific data i.e. after fee and expiry have been removed
-// i.e. The parameter should be the result of DecodeMessage64.
```
fleandrei, line 226: I think this comment should be preserved because this function really expect that the first Uint of it's parameter data to be the operation type.
The only place this function is called is in station-massa-wallet/internal/handler/wallet/sign.go in the getPromptRequest function. It is called with the return value of sendoperation.DecodeMessage64 function. This last function retrieve the fee and the expiry field from the operation payload.
If the data passed to DecodeOperationType is directly the operation payload, it would not work because the first Uint would be the fees. Moreover the DecodeMessage64 function also decode the operation payload from base 64.
I think it would be useful to warn devs about this risk via comments.

peterjah: thx! i did not remember why i removed this ^^but it has nothing do to with this pr ^^

fleandrei (approved):

PR #1388, opened 2025-04-02
use ubuntu-24.04 for ubuntu build

peterjah:


installer/deb/create_deb.sh
```diff
@@ -89,7 +89,8 @@ Maintainer: Massa Labs <massa.net>
 Homepage: https://station.massa.net
 Description: An entrance to the Massa blockchain.
     MassaStation is a secured gateway to the Massa blockchain. This application provides a user-friendly way to access, use and build on the Massa blockchain while keeping you safe from the dangers of the internet.
-Depends: curl, iproute2, libnss3-tools, debconf (>= 0.5) | debconf-2.0
+Depends: debconf (>= 0.5) | debconf-2.0
```
thomas-senechal, line 92: Are we sure we can remove all those dep ? 
For example, `curl` might not be useful anymore, but don't we need `libnss3-tools` and `iproute2` to generate https certificates ? 🤔

.github/workflows/lint.yml
```diff
@@ -4,7 +4,7 @@ on:
 
 jobs:
   lint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
thomas-senechal, line 7: I would put 24.04 just in case
```suggestion
    runs-on: ubuntu-24.04
```

.github/workflows/api.yml
```diff
@@ -10,42 +10,41 @@ jobs:
...
+          sudo apt install -y libwebkit2gtk-4.1-0 xvfb
       - name: Checkout project directory
         uses: actions/checkout@v4
       - name: Install RobotFramework tests dependencies
         working-directory: api/test
         run: |
           pip install -r requirements.txt
-          sudo apt install xvfb
-      - name: Download MassaStation binary
+          # sudo apt install xvfb
```
thomas-senechal, line 28: i guess this line can be deleted as it's not installed in the previous step ?

PR #1403, opened 2025-06-19
Add origin and referer headers

0xB19:


peterjah (approved):
Maybe an issue in the test (see copilot review)

PR #1414, opened 2025-08-06
refactor config

peterjah:
## Delivery Quality Checklist

- [ ] **Changelog:**

- [ ] **PR Dependency:**

- [ ] **Documentation:**


int/api/network/create.go
```diff
@@ -0,0 +1,32 @@
...
+	if body.Default != nil {
+		makeDefault = *body.Default
+	}
+
+	if err := h.configManager.AddNetwork(body.Name, body.URL, makeDefault); err != nil {
+		return operations.NewCreateNetworkBadRequest().WithPayload(&models.Error{Code: "Network-0003", Message: err.Error()})
+	}
+
+	current := h.configManager.CurrentNetwork()
+	response := &models.NetworkManagerItem{CurrentNetwork: &current.Name, AvailableNetworkInfos: []*models.NetworkInfoItem{}}
```
fleandrei, line 30: If AvailableNetworkInfos  is not needed, you may just return the current network name. The endpoint could just return a string rather than a NetworkManagerItem.

If AvailableNetworkInfos is needed, it would be great to provide it.

int/config/network.go
```diff
@@ -1,256 +1,137 @@
...
+	seenNames := make(map[string]string) // lowercase -> original case
 
-	// create the directory if it doesn't exist
-	if _, err := os.Stat(path); os.IsNotExist(err) {
-		err = os.MkdirAll(path, os.ModePerm)
-		if err != nil {
-			return "", fmt.Errorf("creating account directory '%s': %w", path, err)
+	for networkName := range config.Networks {
+		lowerName := strings.ToLower(networkName)
+		if existingName, exists := seenNames[lowerName]; exists && existingName != networkName {
```
fleandrei, line 130: The condition existingName != networkName implies that if there are 2 network with the same name there is no problem. Even if the situation shouldn't occur because of key overwriting during YAML unmarshaling, I think this condition should be removed.

int/config/manager.go
```diff
@@ -0,0 +1,508 @@
...
+	// Remove from in-memory slice
+	idx := -1
+	for i := range n.Network.Networks {
+		if n.Network.Networks[i].Name == name {
+			idx = i
+			break
+		}
+	}
+	if idx != -1 {
+		n.Network.Networks = append(n.Network.Networks[:idx], n.Network.Networks[idx+1:]...)
```
fleandrei, line 496: If idx correspond to the last element of the n.Network.Networks slice, there is a problem

